PR #1, opened 2022-01-06
feat: BlockId latest and pending

milancermak:
This PR adds two new values to the BlockId enum: `Latest` and `Pending`. Most of the API calls, as far as I could figure out, support a `blockId=latest` parameter. Some even work "better" with a `blockId` than for example a `blockNumber` (compare [this one using blockId=36999](https://alpha4.starknet.io/feeder_gateway/get_code?contractAddress=0x052551bca2724c49823effc6e8b0c1084291b832e1c1c195c39671420bff2a6e&blockId=36999) with [this one using blockNumber=36999](https://alpha4.starknet.io/feeder_gateway/get_code?contractAddress=0x052551bca2724c49823effc6e8b0c1084291b832e1c1c195c39671420bff2a6e&blockNumber=36999)) but I haven't discovered it out fully. Need to ask Starkware :)

I've also changed the parameters nomenclature to better represent the value.

Please let me know what you think.

### DIFF
--- a/starknet-core/src/types/block.rs
+++ b/starknet-core/src/types/block.rs
@@ -6,6 +6,8 @@ use serde::Deserialize;
 pub enum BlockId {
     Hash(H256),
     Number(u64),
+    Latest,
+    Pending,
 }
 
 #[derive(Debug, Deserialize)]
--- a/starknet-providers/src/provider.rs
+++ b/starknet-providers/src/provider.rs
@@ -21,22 +21,22 @@ pub trait Provider {
     async fn call_contract(
         &self,
         invoke_tx: InvokeFunction,
-        block_hash_or_number: Option<BlockId>,
+        block_id: Option<BlockId>,
     ) -> Result<CallContractResult, Self::Error>;
 
-    async fn get_block(&self, block_hash_or_number: Option<BlockId>) -> Result<Block, Self::Error>;
+    async fn get_block(&self, block_id: Option<BlockId>) -> Result<Block, Self::Error>;
 
     async fn get_code(
         &self,
         contract_address: H256,
-        block_hash_or_number: Option<BlockId>,
+        block_id: Option<BlockId>,
     ) -> Result<ContractCode, Self::Error>;
 
     async fn get_storage_at(
         &self,
         contract_address: H256,
         key: U256,
-        block_hash_or_number: Option<BlockId>,
+        block_id: Option<BlockId>,
     ) -> Result<U256, Self::Error>;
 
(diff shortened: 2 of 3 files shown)


xJonathanLEI:
Ah I didn't know that. The lack of documentation on the sequencer gateway has always been bugging me.

xJonathanLEI:
@milancermak I did some tests on my side, but it seems that when using `blockId`, the parameter is simply ignored.

With your example:

```sh
$ curl "https://alpha4.starknet.io/feeder_gateway/get_code?contractAddress=0x052551bca2724c49823effc6e8b0c1084291b832e1c1c195c39671420bff2a6e&blockNumber=1"
{"bytecode": [], "abi": {}}
$ curl "https://alpha4.starknet.io/feeder_gateway/get_code?contractAddress=0x052551bca2724c49823effc6e8b0c1084291b832e1c1c195c39671420bff2a6e&blockId=1"
{"bytecode": ["0x40780017fff7fff", "0x1", "0x.....
```

The contract wasn't deployed on block number 1, so it seems that the param is really just ignored.

To make it more obvious, here's an example using `blockId` on `get_block`:

```sh
$ curl "https://alpha4.starknet.io/feeder_gateway/get_block?blockId=1"
{"block_number": 38912, .....
```

This is the same behavior with any other unknown query params, say `asdf`:

```sh
$ curl "https://alpha4.starknet.io/feeder_gateway/get_block?asdf=1"
{"block_number": 38912, .....
```

So I'm afraid `blockId` is actually not supported. We should not code this in.

xJonathanLEI:
That said, it seems that `blockNumber=pending` does work. It gives you the pending block being built:

```sh
$ curl "https://alpha4.starknet.io/feeder_gateway/get_block?blockNumber=pending"
{"status": "PENDING", .....
```

note that this gives you an error when there's "no pending block", whatever it means:

```json
{"code": "StarknetErrorCode.BLOCK_NOT_FOUND", "message": "Pending block does not exist."}
```

xJonathanLEI:
However, despite `blockNumber=pending` working, my sense is that due to the lack of formal official documentation, we should take the client implementation in `cairo-lang` as if it's the official documentation. This way we won't make any more breaking changes than those of `cairo-lang` itself.

So I guess we will have to wait until this is published in `cairo-lang` before we build this in.

milancermak:
Thanks for double checking, I guess I was too eager and not thorough enough with my testing. 100% agree that we should mimic what's in `cairo-lang`. Can't remember where I found the `pending` block identifier now...

PR #2, opened 2022-01-06
refactor: using rename_all

milancermak:
Just a small thing, using serde's [rename-all](https://serde.rs/container-attrs.html) in a couple of places. I think it makes the code more readable.


xJonathanLEI (approved):
LGTM

PR #3, opened 2022-01-07
fix: use symbolic links for files in `assets` folder

xJonathanLEI:
This PR changes to use symbolic links for files in `./assets`.

`README.md` files from sub-crates are linked to a top-level `assets` folder since sub-crate folders are excluded when packaging `starknet`. However, hard links were incorrectly used, which are not supported by Git.


xJonathanLEI:
I'm merging this directly since no actual Rust code change is made.

Issue #4, opened 2022-01-07
Write tests and set up CI pipeline for running them


xJonathanLEI:
There's currently zero test written for the project, making it hard to contribute without breaking stuff. We need to write tests and run them on `master` and PRs. After this is done, all new PRs that add/change Rust code must be accompanied with corresponding test cases.

TODOs:

- [x] add basic tests
- [x] set up CI for running tests
- [x] set up CI for `audit-check`, `cargo-udeps`, and `cargo-deny`

milancermak:
I just started to work on a basic unit test suite 🤘I'll take this one then.

milancermak:
@xJonathanLEI I've set up a simple [test action](https://github.com/xJonathanLEI/starknet-rs/actions/workflows/test.yaml) in 20f886bd - it runs `cargo test` for stable and nightly on the project. It's the first time I'm creating this for a Rust project so please let me know if I missed anything.

While researching how to do this, I've stumbled upon an [audit check](https://github.com/actions-rs/audit-check) action. Seems useful, should I set it up? And since I'm on the topic, what do you think about having [cargo udeps](https://github.com/est31/cargo-udeps) and [cargo deny](https://github.com/EmbarkStudios/cargo-deny)?

xJonathanLEI:
@milancermak Nice one on the PR! I'm gonna check in a bit.

As for `cargo-udeps` and `cargo-deny`, I think it's a good idea to set those up as part of CI. But I'd probably skip the license check in `cargo-deny`, as we're not distributing bundled binaries anyways, and I'm not really familiar with license compatibility and stuff lol.

milancermak:
Aye, good point on the license check in `cargo-deny`.

Will set up the rest tomorrow and open a PR.

xJonathanLEI:
Closing as all items have been done.

PR #7, opened 2022-01-07
Adding basic tests

milancermak:
Basic unit tests for core types. Linked to #4.

TODO:
 * [x] full type coverage
 * [x] set up a test Action
 * [x] import fixtures with `include_str`


milancermak:
When working on the test for `ContractCode`, I borrowed the [test fixture from `cairo-lang` repo](https://github.com/starkware-libs/cairo-lang/blob/3d33c4e829a87bc3d88cf04ed6a489e788918b8b/src/starkware/starknet/compiler/starknet_preprocessor_test.py#L143) and noticed that in one of the examples of a function, there's a `"stateMutability": "view"` field. It [seems](https://github.com/starkware-libs/cairo-lang/blob/3d33c4e829a87bc3d88cf04ed6a489e788918b8b/src/starkware/starknet/compiler/starknet_preprocessor.py#L129) the attribute does only have `view` as its value and appears only if a Cairo function is a `@view`. The type doesn't yet have this attribute, I'll add it later.

xJonathanLEI:
One comment: for raw responses, shall we put them in a test vector folder and import them thru [include_str](https://doc.rust-lang.org/std/macro.include_str.html)? Should be good for readability. Also `rustfmt` seems to [have issues](https://github.com/rust-lang/rustfmt/issues/3863) formatting long lines.

milancermak:
Good idea.

You are more experienced in Rust than me, where should the folder be? It seems to me there's not a consensus amongst the Rust community where to place test fixtures (and TBH a lack of great test tooling, but that's a different issue). Normally, in e.g. python, I'd have them under `tests/fixtures`, but in Rust `tests/` is, by convention, for integration tests. What's your take?

xJonathanLEI:
I agree that there doesn't seem to be a de-facto way of putting test vectors. Maybe let's just agree to put them in a `test-data` folder alongside `src`, under the sub-crate being tested? For example, if we're testing `starknet-core`, then it should be like this:

```
.
├── Cargo.toml
└── starknet-core
    ├── src
    │   ├── lib.rs
    │   └── types
    │       └── mod.rs
    └── test-data
        └── raw_gateway_responses
            ├── get_block
            │   ├── raw_response_1.txt
            │   ├── raw_response_2.txt
            │   └── raw_response_3.txt
            ├── get_code
            │   ├── raw_response_1.txt
            │   ├── raw_response_2.txt
            │   └── raw_response_3.txt
            └── get_contract_addresses
                ├── raw_response_1.txt
                ├── raw_response_2.txt
                └── raw_response_3.txt
```

It doesn't really make a difference, but for generating those raw response files, we should probably use the `--output` flag of `curl` instead of directing the standard output with `>` to keep the file _exactly_ the same as the data sent over the wire (without the extra empty line).

Also we should use the `txt` extension instead of `json` for the raw response files to avoid getting format-checked.

milancermak:
Crap, I think I messed up a rebase and now the commits from `master` appear here. Gonna force-push a fix to have it clean.

milancermak:
I think this covers all the types except for `CallContractResult` which is pretty simple but I'm too tired now to go through all the hurdles of mimicking a POST request to `/call_contract` to get a raw response 😇 so I hope it's fine without it.

xJonathanLEI:
@milancermak It seems that 0 test was run in the CI pipeline you set up, presumably `cargo test` was run instead of `cargo test --all`.

xJonathanLEI:
Regarding the fake JSON response at `starknet-core/test-data/raw_gateway_responses/get_code/2_all_abi_types.txt`, I've replaced it with a real one by actually [deploying](https://alpha4.starknet.io/feeder_gateway/get_code?contractAddress=0x06ef97a90be1c0458f6e7bd1faf05021f2d81211f658155df0c5c97a39eb2d12) the said contract onto `alpha-goerli`.

xJonathanLEI:
I've also did a small refactoring by splitting test cases where multiple raw responses are loaded into separate cases.

milancermak:
Cheers 🥳

PR #8, opened 2022-01-07
Add format checking for md, yaml, and json files

xJonathanLEI:
This PR adds formatting checking for the following file types on top of `.rs` files:

- `md`
- `yaml` and `yml`
- `json`

`jq` is used for `json` files instead of `prettier` since `prettier` seems to accept multiple representations of the same `json` file in some cases.


xJonathanLEI:
Again, merging directly since there's no Rust code change.

PR #10, opened 2022-01-07
feat: get_last_batch_id

milancermak:
The feeder gateway client in `cairo-lang` has a [`get_last_batch_id` function](https://github.com/starkware-libs/cairo-lang/blob/master/src/services/everest/api/feeder_gateway/feeder_gateway_client.py#L17) that's missing in the lib. This PR adds it. The call itself [returns](https://alpha4.starknet.io/feeder_gateway/get_last_batch_id) a single number.


xJonathanLEI:
You're right. I missed the based class.

xJonathanLEI (approved):
LGTM

Issue #6, opened 2022-01-07
Support smart contract deployment


xJonathanLEI:
Now that there's full sequencer gateway API coverage, smart contract deployment can be built. This issue tracks the progress of contract deployment support.

TODOs:

- [x] figure out what the `program` field in deployment requests represents
- [x] contract artifact parsing and `program` compression
- [x] add type for deploying artifact through `starknet_providers::Provider`

xJonathanLEI:
After looking into the [code in cairo-lang](https://github.com/starkware-libs/cairo-lang/blob/7712b21fc3b1cb02321a58d0c0579f5370147a8b/src/starkware/starknet/services/api/gateway/transaction.py#L63), it's clear that the `program` field under `contract_definition` in a contract deployment request is simply the base-64 representation of the gzipped JSON text of the `program` field of a contract artifact.

Basically, to get the compressed `program` from a contract compilation artifact:

```sh
$ cat /path/to/contract/artifact.json | jq -c ".program" | gzip | base64 -w 0
```

xJonathanLEI:
The `program` field of a compiled contract artifact has 8 fields:

- builtins
- data
- debug_info
- hints
- identifiers
- main_scope
- prime
- reference_manager

of which `debug_info` is _not_ required for a `deploy` transaction, despite the official `cairo-lang` putting it inside the compressed representation.

xJonathanLEI:
All tasks done. The library can now be used for smart contract deployment. Closing issue.

PR #19, opened 2022-01-10
feat: implement `ContractFactory` for contract deployment

xJonathanLEI:
Resolves task 3 of #6:

> add type for deploying artifact through `starknet_providers::Provider`

### DIFF
--- a/starknet-contract/src/factory.rs
+++ b/starknet-contract/src/factory.rs
@@ -1,0 +1,80 @@
+use crate::ContractArtifact;
+
+use flate2::{write::GzEncoder, Compression};
+use rand::{thread_rng, RngCore};
+use starknet_core::types::{
+    AbiEntry, AddTransactionResult, ContractDefinition, DeployTransactionRequest,
+    EntryPointsByType, TransactionRequest, H256, U256,
+};
+use starknet_providers::Provider;
+use std::io::Write;
+
+pub struct Factory<P>
+where
+    P: Provider,
+{
+    compressed_program: Vec<u8>,
+    entry_points_by_type: EntryPointsByType,
+    abi: Vec<AbiEntry>,
+    provider: P,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum FactoryError {
+    #[error(transparent)]
+    CannotSerializeProgram(serde_json::Error),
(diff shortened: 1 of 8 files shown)


xJonathanLEI:
Note that this PR adds an integration test for deploying a contract to `alpha-goerli`. To avoid unnecessarily spamming the network, when developing locally, you're advised to skip running the test case unless it's relevant.

starknet-contract/src/factory.rs
```diff
@@ -0,0 +1,80 @@
...
+        rng.fill_bytes(&mut salt_buffer[1..]);
+
+        self.provider
+            .add_transaction(
+                TransactionRequest::Deploy(DeployTransactionRequest {
+                    contract_address_salt: H256::from(salt_buffer),
+                    contract_definition: ContractDefinition {
+                        program: self.compressed_program.clone(),
+                        entry_points_by_type: self.entry_points_by_type.clone(),
+                        abi: Some(self.abi.clone()),
```
milancermak, line 72: Could there actually be a contract without an ABI?

milancermak (approved):
All good

starknet-contract/src/factory.rs
xJonathanLEI: Yes. Just manually change this to `None` and it'll still work.

Also as noted [here](https://community.starknet.io/t/remove-abi-from-contract-deployment-request-and-get-code-response/308/2), there's really no way for the sequencer to validate the ABI. I think that's why they made it optional.

I was thinking about whether we should give users a choice on this too, by maybe making `abi` optional on the artifact type. What do you think?

xJonathanLEI: So maybe we should expose the compression logic on the `Program` type, and have users call the provider directly if they want to skip ABI?

Actually, I was initially going to build the compression logic there. Don't remember why I ended up putting in the factory ctor, maybe too sleepy lol.

xJonathanLEI:
Anyways. Lemme merge this first and refactor later on if needed :)

starknet-contract/src/factory.rs
milancermak: I have no opinion on this, but a cool hack on deploying the fake ABI 🎩

PR #20, opened 2022-01-10
ci: use rust-cache for udeps

### LINKED ISSUE #17: `cargo-udeps` CI pipeline takes too long
Follow up on #16. As @milancermak noted, the udeps pipeline is taking quite long to run.

milancermak:
Resolves #17 

I changed the cache step for the same as is used in the Test action, udeps [now take](https://github.com/xJonathanLEI/starknet-rs/runs/4768223668?check_suite_focus=true) about a minute 🥳

### DIFF
--- a/.github/workflows/udeps.yml
+++ b/.github/workflows/udeps.yml
@@ -19,14 +19,9 @@ jobs:
           profile: minimal
           override: true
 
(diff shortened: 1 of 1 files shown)


xJonathanLEI (approved):
Good one

PR #22, opened 2022-01-11
docs: add contract deployment example to README

xJonathanLEI:
Now that #6 is done and contract deployment is fully working, we can update the feature checkbox in `README.md`.

This PR also adds an example code snippet for reading an artifact from file and deploying to testnet.


milancermak (approved):
🚢

PR #24, opened 2022-01-12
ci: run tests on OS X and Windows

### LINKED ISSUE #23: Run tests on OS X and Windows
Since we're building a library, we want to make sure it is available on all mainstream dev platforms. We're currently running the test pipeline only on Linux. Add OS X and Windows too.

milancermak:
Resolves #23.

Going from stable & nightly on Ubuntu to 1) stable & nightly on Ubuntu, 2) stable and nightly on OS X 11, 3) stable & nightly on Win GNU and 4) stable and nightly on Win MSVC.

This way we probably cover all the major dev platforms.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,14 +6,13 @@ on:
 
 name: Tests
 jobs:
-  test:
-    name: Rust tests
-    runs-on: ubuntu-latest
+  unix-test:
+    name: Unix tests
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        rust:
-          - stable
-          - nightly
+        os: [ubuntu-latest, macos-11]
+        toolchain: [stable, nightly]
 
     steps:
       - name: Checkout source code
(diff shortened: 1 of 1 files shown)


milancermak:
This [initial run](https://github.com/xJonathanLEI/starknet-rs/actions/runs/1689743489) was rather long, but I think that's because there was no cache hit, subsequent runs should be faster.

xJonathanLEI (changes requested):
In the test command you're missing the `--target` flag on Windows, which results in `msvc` always being tested on both variants. Using `actions-rs/toolchain@v1` won't automatically set that up.

xJonathanLEI:
Not sure if that applies to `stable`/`nightly` too. You might have to use `cargo +nightly test` on nightly channels. For this one I haven't tested yet though.

The `--target` one, on the other hand, has indeed been tested, and explicitly specifying it is needed.

xJonathanLEI (approved):
Nice job 🚀

milancermak:
Cheers, and thank you for spotting the error in the first place.

Issue #27, opened 2022-01-18
Add support for StarkNet v0.7.0


xJonathanLEI:
Support [StarkNet v0.7.0](https://github.com/starkware-libs/cairo-lang/releases/tag/v0.7.0).

TODOs:

- [x] pending block query
- [x] events in transaction receipts
- [x] resolves breaking changes
- [x] reading contract artifacts with events

xJonathanLEI:
While pending block query is live on both `alpha-goerli` and `alpha-mainnet`, events seem to be on `alpha-goerli` at the moment.

If we simply add `events` to transaction receipts, the client will fail on `alpha-mainnet`. Making the field optional seems just wrong to me: it really isn't optional in a up-to-date network, after all.

Generally speaking, how should we handle situations like this?

@milancermak Any thoughts?

xJonathanLEI:
Another option is to review and approve the branch, but not merge it into `master` until `alpha-mainnet` is updated to support said features.

xJonathanLEI:
Oops. Funny. They **removed** the `status` field of transaction receipts in `get_block` responses, which breaks our provider on `alpha-goerli` because we defined that field as optional.

Let me check with StarkNet to see if this is part of the v0.7.0 upgrade. If so we need to also add it to the TODO list.

xJonathanLEI:
If we take the "not merging to `master` until on mainnet" approach, we should probably maintain a "testnet" branch, which is strictly not-behind `master` and contains all feature additions and removals.

But there's still a problem. We can't possibly write our tests (especially integration tests) against mainnet. So we'll effectively still be running `master` tests against testnet, but the problem is `master` isn't supposed to be compatible with testnet anymore.

xJonathanLEI:
Well... in this very case I'm gonna just request the removal of those fields into `master` anyways, but the discussion above on the management of feature addition/removal still stands. Maybe we should turn that into a GitHub Discussion instead.

xJonathanLEI:
One more breaking change found. Before `v0.7.0` calling `get_transaction_receipt` on non-existent transaction hashes result in an object containing the hash passed in as `transaction_hash`. Now it's just `0x0`.

milancermak:
What about having `master` in sync with the testnet (`alpha-goerli`) and whatever is  cargo released working with mainnet?

[...]

PR #29, opened 2022-01-19
feat: resolve breaking changes from StarkNet 0.7.0

xJonathanLEI:
Resolves task 3 of #27:

> resolves breaking changes

Specifically, the breaking changes are:

1. The fields `block_hash`, `block_number`, and `status` are removed from `transaction_receipts` elements in `get_block` responses.
2. `get_transaction_receipt` requests for non-existent hashes now return with `0x0` as `transaction_hash`.

This PR:

- adds a `generate_responses.sh` script for documenting how test data were generated for gateway responses;
- updates `raw_gateway_responses` test data against the current `alpha-goerli` gateway;
- resolves the breaking changes listed above.


xJonathanLEI:
Note that it seems that the issue of the randomly shuffling of JSON fields on gateway responses has been resolved. We should have more deterministic test data generation going forward (no more different JSON representation for the exact same thing with the only difference being the ordering of fields).

milancermak (approved):
Nice work!

PR #32, opened 2022-01-20
feat: add support for pending block query

xJonathanLEI:
Resolves task 1 of #27:

> pending block query


starknet-core/src/types/block.rs
```diff
@@ -80,10 +86,22 @@ mod tests {
...
+        assert_eq!(block.block_number.unwrap(), 47543);
         assert_eq!(block.transaction_receipts.len(), 4);
         let receipt = &block.transaction_receipts[3];
         assert_eq!(receipt.events.len(), 1);
         assert_eq!(receipt.events[0].data.len(), 2);
     }
+
+    #[test]
+    fn test_block_deser_pending() {
+        // has an L2 to L1 message
```
milancermak, line 98: The comment doesn't make sense in this context, probably a copy pasta err?

starknet-core/src/types/block.rs
```diff
@@ -1,23 +1,29 @@
...
+    Pending,
 }
 
 #[derive(Debug, Deserialize)]
 pub struct Block {
-    #[serde(deserialize_with = "deserialize_h256_from_hex")]
-    pub block_hash: H256,
-    pub block_number: u64,
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_option_h256_from_hex")]
```
milancermak, line 18: A very minor nit - the `default` and `deserialize_with` could be written on a single line, in one attribute. Same applies to the `state_root` below. I'm fine with merging as it tho.

milancermak (approved):
All good, I only left two nitpicky comments.

starknet-core/src/types/block.rs
xJonathanLEI: Yes copy pasta got caught

starknet-core/src/types/block.rs
xJonathanLEI: Yes ser changing it

PR #37, opened 2022-01-25
feat: implement RFC 6979 for deterministic ephemeral scalar generation

### LINKED ISSUE #34: Implement RFC 6979 for deterministic `k` generation
After #33 is merged, we'll need to implement [RFC 6979](https://datatracker.ietf.org/doc/html/rfc6979#section-3) before being able to build high-level crypto functions into `starknet-core`.

Technically, we don't _have to_ use RFC 6979 to generate valid signatures given strong enough entropy. But we should still use it for the sake of behaving the same as other libraries and generating deterministic signatures.

xJonathanLEI:
Resolves #34.

With the RFC 6979 support we'll be able to generate the exact same signatures as `cairo-lang` and a couple of other StarkNet libraries.


starknet-crypto/src/rfc6979.rs
```diff
@@ -0,0 +1,121 @@
...
+pub fn generate_k(
+    message_hash: &FieldElement,
+    private_key: &FieldElement,
+    seed: Option<&FieldElement>,
+) -> FieldElement {
+    // The message hash padding as implemented in `cairo-lang` is not needed here. The hash is
+    // padded in `cairo-lang` only to make sure the lowest 4 bits won't get truncated, but here it's
+    // never getting truncated anyways.
+    let message_hash = U256::from_be_slice(&message_hash.to_repr().0).to_be_byte_array();
+    let private_key = crypto_bigint::U256::from_be_slice(&private_key.to_repr().0);
```
milancermak, line 27: nit: `U256` doesn't need to be fully qualified

milancermak (approved):
Great stuff ✨

starknet-crypto/src/rfc6979.rs
xJonathanLEI: Yes you're right. Missed it when cleaning up junk code lol.

Issue #13, opened 2022-01-09
Replace `H256` and `U256` with `StarkHash` and `StarkInteger`


xJonathanLEI:
For convenience, the library currently uses `H256` and `U256` from the `ethereum_types` crate to represent hashes and integers on StarkNet. However, the range of these hashes and integers are actually smaller than 256 bits. Exposing public interfaces with `H256` or `U256` is thus misleading.

We should implement the `StarkHash` and `StarkInteger` types (names open to discussion) to replace the existing Ethereum types used, probably as wrappers around `H256` and `U256`, respectively, with range checks.

milancermak:
I think they might be [felts](https://www.cairo-lang.org/docs/how_cairo_works/cairo_intro.html#field-elements), which is a 252-bit number, but doesn't quite need all the 252 bit space, as max value is 2**251 + 17 * 2**192 + 1.

xJonathanLEI:
Yes they're felts, so we better name then `FieldElement` instead of `StarkInteger`? Anyways we still need something like that with range check.

BTW I forgot that these numbers can be negative too, so I used `U256` to represent them. Need to fix.

milancermak:
Personally, I got used to the name `felt`, but `FieldElement` is fine with me.

I don't think the values can be negative, just in the context of Cairo, negative numbers are represented as the "upper half" of the felt's value. At least that's my understanding.

xJonathanLEI:
I think you're right that they're only negative in the context of Cairo. Aka. they're not really two's-complement negative. I'm not entirely sure and will need to do some tests on this.

But imaging having contract bindings (we will pretty soon after implementing a signer) that return `felt`s, users would definitely expect a negative value under Cairo context to be negative in the client code too.

xJonathanLEI:
How about this? We treat literally ALL values (including hashes and contract code and call contract results) as a single type: `UnsignedFieldElement`, which can then be converted into `SignedFieldElement` if users choose to. In this case, we'll use `SIgnedFieldElement` for return types in `abigen` binding code, for the sake of fitting into the Ciaro context.

The reason is that it seems the boundary between hashes and numbers are quite blurry in StarkWare's context. They're literally the same thing after all. So we might as well just give up on separating them.

milancermak:
Great idea ❤️ I think that will work nicely.

xJonathanLEI:
OK I'm starting to work on this now. The issue has become quite burning as our API surface grows.

[...]

PR #38, opened 2022-01-26
feat: implement high-level Pedersen util in `starknet-core`

xJonathanLEI:
This PR adds a `compute_hash_on_elements` utility function to `starknet-core`. This function is useful for computing the sig hash for the account contract.

Code ported from #18 with the underlying crypto library replaced with `starknet-crypto`.

A new public function `to_bytes_le` was added to `starknet_crypto::FieldElement` to make this possible.

### DIFF
--- a/starknet-core/src/crypto.rs
+++ b/starknet-core/src/crypto.rs
@@ -1,0 +1,83 @@
+use ethereum_types::{H256, U256};
+use starknet_crypto::{pedersen_hash, FieldElement};
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum PedersenHashError {
+    #[error("data must not be empty")]
+    EmptyData,
+    #[error("element out of range: {0}")]
+    ElementOutOfRange(U256),
+}
+
+pub fn compute_hash_on_elements(data: &[U256]) -> Result<H256, PedersenHashError> {
+    if data.is_empty() {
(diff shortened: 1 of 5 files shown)


starknet-core/src/crypto.rs
```diff
@@ -0,0 +1,83 @@
...
+    let data_len = U256::from(data.len());
+    current_hash = pedersen_hash(
+        &current_hash,
+        &u256_to_field_element(&data_len).ok_or(PedersenHashError::ElementOutOfRange(data_len))?,
+    );
+
+    Ok(H256::from_slice(&current_hash.to_bytes_le()[..]))
+}
+
+fn u256_to_field_element(num: &U256) -> Option<FieldElement> {
```
milancermak, line 37: Just curious here - wouldn't it be cleaner (which is subjective) if this function lived in `field_element.rs` in `starknet-crypto`?

milancermak (approved):
LGTM 👏

starknet-core/src/crypto.rs
xJonathanLEI: `starknet-crypto` isn't aware of `U256`. The `U256` in `starknet-crypto` is `crypto_bigint::U256` instead of the `primitive_types::U256` here, despite the same name lol.

Actually we should be getting rid of `U256` and `H256` altogether... as suggested in #13.

PR #42, opened 2022-01-27
feat: add `SigningKey` and `VerifyingKey` for local key pair operations

xJonathanLEI:
This PR:

- adds a new sub-crate `starknet-signers`;
- adds a new type `KeyPair` which further wraps crypto functionalities previously introduced in `starknet-core`.


xJonathanLEI:
Oops I mistakenly escaped the backticks in the PR title LOLLL. Thought it was the command line.

xJonathanLEI:
A generic `Signer` trait will be introduced later on, with `KeyPair` being an implementation.

The `Signer` trait will only expose `async` methods though, as really anything other than an in-memory private key does crypto operations asynchronously (think Ledger).

starknet-signers/src/key_pair.rs
```diff
@@ -0,0 +1,136 @@
+use starknet_core::{
+    crypto::{ecdsa_sign, EcdsaSignError, Signature},
+    types::{H256, U256},
+};
+use starknet_crypto::{get_public_key, FieldElement};
+
+#[derive(Debug)]
+pub struct KeyPair {
+    private_key_u256: U256,
+    private_key_fe: FieldElement,
```
milancermak, line 10: I don't like this struct too much, because they are both representation so the same and one is derived from the other. Since it's a key *pair*, what about having both `private_key` and `public_key` as pub members of the struct? 
```rust
pub struct KeyPair {
    pub private_key: U256,
    pub public_key: U256
}
```

Granted, one can still be derived from the other, but at least they're different "category".

Or alternatively have a `Secret` instead of a `KeyPair` which would wrap only the private key?

xJonathanLEI: Hmm I think having the public key computed by default is kind of weird... What if the user never calls `get_public_key`? That computing would be wasted right?

xJonathanLEI: I also don't like having 2 fields of literally the same thing here... But it's opaque to users as they're private. It's there for performance reasons just in case it's not obvious.

xJonathanLEI: Maybe we change the name to `PrivateKey` or `SigningKey`? The equivalent in `ethers-rs` seems to be named `SigningKey`, and the public key type is called `VerifyingKey`.

xJonathanLEI: Can also remove the duplicated fields to make it more idiomatic, though it's not visible to users.

milancermak: Yeah, it's wasted in this sense. Do you think it's too much?

BTW it's true for `private_key_fe` as well - it's only used in `get_public_key`, though converting `U256` to `FieldElement` isn't as expensive.

xJonathanLEI: Well that isn't totally wasted as the process of converting to `FieldElement` in `new()` is also responsible for the range check.

milancermak: I like that solution of `SigningKey` and `VerifyingKey` like in `ethers-rs`.

xJonathanLEI: Cool. Lemme refactor it a bit.

xJonathanLEI: There you go. Refactored.

xJonathanLEI:
I have to admit all these stupid range checks are getting really annoying. Looks like it's time to pick up #13 and unify all of them for good.

A lot of these checks should be eliminated once that's done, as the type itself is already automatically bounded to the said range.

milancermak (approved):
Yay 🎆

starknet-core/src/crypto.rs
```diff
@@ -169,8 +197,52 @@ mod tests {
...
+            .unwrap();
+
+        assert_eq!(
+            ecdsa_verify(&public_key, message_hash, &Signature { r, s }).unwrap(),
+            true
+        );
+    }
+
+    #[test]
+    fn test_ecdsa_verify_invvalid_signature() {
```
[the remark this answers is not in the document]
xJonathanLEI: Yes ser

PR #44, opened 2022-01-28
refactor: replace `H256` and `U256` with `UnsignedFieldElement` in VO's

xJonathanLEI:
Resolves task 2 of #13:

> replace `U256` and `H256` with `UnsignedFieldElement` in value objects and providers

This is a messy one, as we're fixing all the misuse of `H256` and `U256` across a lot of files.


starknet-core/src/types/transaction.rs
```diff
@@ -26,8 +28,8 @@ pub struct TransactionWithStatus {
     pub transaction: Option<Transaction>,
     pub status: TransactionStatusType,
     #[serde(default)]
-    #[serde(deserialize_with = "deserialize_pending_block_hash")]
-    pub block_hash: Option<H256>,
+    #[serde(deserialize_with = "pending_block_hash_de")]
```
milancermak, line 31: The `#[serde]`s here could be merge into one line.

milancermak (approved):
Looking good!

Kinda shame that the API uses both decimal and hex representations for the FieldElement values.

starknet-core/src/types/transaction.rs
xJonathanLEI: Yes ser

xJonathanLEI:
Yeah should have just followed Ethereum and make everything hex.

PR #45, opened 2022-01-28
refactor: remove `U256` and `H256` on all public APIs

xJonathanLEI:
Resolves task 3 of #13:

> replace `U256` and `H256` with `UnsignedFieldElement` in crypto functions


milancermak (approved):
the changes to the `key_pair.rs` 😍

PR #47, opened 2022-01-28
feat: add `Signer` trait

xJonathanLEI:
Resolves task 1 of #50:

> add `Signer` trait as abstraction over `SigningKey` and other future implementations

This PR:

- adds a new `Signer` trait;
- adds a new type `LocalWallet` which wraps `SigningKey` and implements `Signer`.

In `ethers-rs`, the `address` method is synchronous, forcing all implementations that are asynchronous in nature to pre-fetch the address in their constructor, even if the user doesn't want to know the address. I took a different approach and made it `async` and free implementations from eagerly loading the address.

### DIFF
--- a/starknet-signers/src/local_wallet.rs
+++ b/starknet-signers/src/local_wallet.rs
@@ -1,0 +1,49 @@
+use crate::{Infallible, Signer, SigningKey};
+
+use async_trait::async_trait;
+use starknet_core::{
+    crypto::{EcdsaSignError, Signature},
+    types::UnsignedFieldElement,
+};
+
+pub struct LocalWallet {
+    private_key: SigningKey,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum SignError {
+    #[error(transparent)]
+    EcdsaSignError(EcdsaSignError),
+}
+
+impl LocalWallet {
+    pub fn from_signing_key(key: SigningKey) -> Self {
+        key.into()
+    }
+}
+
+#[async_trait]
+impl Signer for LocalWallet {
+    type GetAddressError = Infallible;
+    type SignError = SignError;
+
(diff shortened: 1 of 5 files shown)


xJonathanLEI:
This trait is subject to change. It currently exposes a (dangerous) `sign_hash` method, which I reasonably assume will not be supported by hardware wallets. This is the best we can do right now before any such wallets add StarkNet support though.

xJonathanLEI:
`LocalWallet` looks pretty pointless for now. Why not just implement `Signer` for `SigningKey`?

Well, once we replace `sign_hash` with something equivalent to `sign_transaction` in `ethers-rs`, this meaningless-looking wrapping will make sense as this wrapper will then handle the logic of converting transaction data into a hash.

xJonathanLEI:
Made a mistake. That shouldn't be called `get_address`. We have account abstraction and "address" should mean the account contract address. Changed to `get_public_key` instead.

starknet-signers/src/local_wallet.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl LocalWallet {
+    pub fn from_signing_key(key: SigningKey) -> Self {
+        key.into()
+    }
+}
+
+#[async_trait]
+impl Signer for LocalWallet {
+    type GetPublicKeyError = Infallible;
```
milancermak, line 27: `Infallible` exists just to comply with the API and as to mark that this one can never fail?

milancermak (approved):
This is looking good, way to go.

starknet-signers/src/local_wallet.rs
xJonathanLEI: Yep just to comply with the `Signer` API, as that's a generic one and other implementations will be fallible due to a lot of reasons. This is similar to `std::convert::Infallible`.

PR #48, opened 2022-01-29
fix: fix deserializer used for contract call result

xJonathanLEI:
The `result` field in `CallContractResult` uses hex representation. This bug wasn't discovered during the `UnsignedFieldElement` refactor because there's no test case covering this functionality.


milancermak (approved):
LGTM

PR #49, opened 2022-01-29
fix: fix Pedersen hash on empty slice

xJonathanLEI:
Just checked the official `cairo-lang` implementation and turns out you can do `compute_hash_on_elements` on an empty array.

This PR fixes that. Now `compute_hash_on_element` is infallible and returns `UnsignedFieldElement` instead of `Result`.


milancermak (approved):
Awesome

Issue #50, opened 2022-01-29
Implement wrapper around the standard `IAccount` interface


xJonathanLEI:
As set out in our feature roadmap, we need to support the [IAccount](https://github.com/OpenZeppelin/cairo-contracts/blob/main/contracts/IAccount.cairo) interface for using account abstraction.

TODOs:

- [x] add `Signer` trait as abstraction over `SigningKey` and other future implementations
- [x] add `Account` trait which models [IAccount](https://github.com/OpenZeppelin/cairo-contracts/blob/main/contracts/IAccount.cairo), and an implementation for the single-owner variant
- [x] add example to `README` and mark feature as done :)))

xJonathanLEI:
I already have a working implementation for single-owner accounts, but I'm waiting for #47 to be merged first before submitting as it depends on that.

xJonathanLEI:
All tasks done. Closing issue.

PR #51, opened 2022-01-30
feat: add `Account` trait with single-owner implementation

xJonathanLEI:
Resolves task 2 of #50:

> add `Account` trait which models [IAccount](https://github.com/OpenZeppelin/cairo-contracts/blob/main/contracts/IAccount.cairo), and an implementation for the single-owner variant


milancermak:
Exciting! I'm having a look, but being "distracted" by dad duties. Will finish the review today for sure though.

xJonathanLEI:
No rush dude. Take your time!

starknet-accounts/src/single_owner.rs
```diff
@@ -0,0 +1,139 @@
...
+
+        let mut execute_calldata = vec![to, selector, calldata.len().into()];
+        for item in calldata.iter() {
+            execute_calldata.push(*item);
+        }
+        execute_calldata.push(nonce);
+
+        self.provider
+            .add_transaction(
+                TransactionRequest::InvokeFunction(InvokeFunctionTransactionRequest {
```
milancermak, line 128: Outside of the scope of this PR, but the renaming of `InvokeFunctionTransaction` to `InvokeFunctionTransactionRequest` got me really confused when I was navigating the code (I was also looking at the `InvokeFunction` in call_contract.rs which has the same structure, so I was like 🤯 ). Is it common in Rust to rename types like this in `mod.rs`?

starknet-accounts/src/single_owner.rs
```diff
@@ -0,0 +1,139 @@
...
+            .map_err(ExecuteError::SignerError)?;
+
+        let mut execute_calldata = vec![to, selector, calldata.len().into()];
+        for item in calldata.iter() {
+            execute_calldata.push(*item);
+        }
+        execute_calldata.push(nonce);
+
+        self.provider
+            .add_transaction(
```
milancermak, line 127: Why the use of `add_transaction` instead of `call_contract`?

starknet-accounts/src/account.rs
```diff
@@ -0,0 +1,22 @@
+use async_trait::async_trait;
+use starknet_core::types::{AddTransactionResult, BlockId, UnsignedFieldElement};
+use std::error::Error;
+
+#[async_trait]
+pub trait Account {
+    type GetNonceError: Error + Send;
```
milancermak, line 7: From a user perspective, I find it weird to have different error types per `IAccount` functions. What's the advantage compared to having just a single error enum? If OZ extends the API and adds a new external function, we would have to add an error type here right?

milancermak (commented):
It's getting late here, I didn't get to actually use the code, so I just have a couple of questions so far.

xJonathanLEI:
Thanks for the hard work!

starknet-accounts/src/account.rs
xJonathanLEI: There are a few options here off my head:

- Exactly what I'm doing here: passing in error types as generic parameters
- The trait itself enforces an error type, which in turn wraps a custom inner error type supplied through generic param. Essentially the same stuff as option 1
- Same as option 2 but wraps a `&'static str` instead

Out of the 3 options, only option 3 is "interportable". `serde_json` uses it AFAIK.

Since it's obviously impossible to create an umbrella error type that covers every possible future signer error scenario without resorting to static strings or magic integers (remember, adding enum variants is considered a source breaking change in Rust), I think it really comes down to whether it's expected that users would process the errors they get, or they'll just print to log and forget about it, since structured error types are much easier to handle than strings or magic codes.

So I made the design decision to use option 1. I also looked at `ethers-rs` (as always), and they seem to be doing the same for error types.

xJonathanLEI: Oh I think I misunderstood your comment lol. You're talking about per-funciton and not per-implementation.

xJonathanLEI: OK so for the per-function error type, I think it also comes down to whether it's expected that users will handle errors based on type. It's obviously impossible to get a signer-related error when calling `get_nonce`, but if we use a single error enum, users would have to either:

- unnecessarily list all possible error types in `match` arms and discard; or
- use a catch-all discard (a bit of code smell)

I'm not sure if this is the ideal design here. An alternative would be to have a single umbrella enum with each variant corresponding to one "type" of functions. But in that case, adding a new function to the trait WILL be a breaking change (since you're adding enum variants.

starknet-accounts/src/single_owner.rs
xJonathanLEI: `add_transaction` is the equivalent to `eth_sendRawTransaction` and `call_contract` is to `eth_call`.

starknet-accounts/src/single_owner.rs
xJonathanLEI: I think it's common in Rust to name types as simple as possible _inside_ modules. For example you have `std::io::Error` not `std::io::StdIoError`. When re-exporting, clashes are inevitable due to this convention.

Sometimes it's not even about clashes, but rather that getting re-exported to an upper level means it loses the context provided by the module path, so you have to re-name. See `ethers-rs` for a lot of renaming in re-exports. For one, `Wallet` is exported as `LocalWallet`, etc.

starknet-accounts/src/single_owner.rs
milancermak: D'oh, of course.

starknet-accounts/src/account.rs
milancermak: Makes sense. The argument of breaking for less people when using separate enums is solid. I imagine in a single enum type scenario, people would be using a catch-all more which would lead to things breaking silently.

milancermak: Thanks for the discussion as well!

milancermak (approved):

starknet-accounts/src/account.rs
xJonathanLEI: Yeah the thing I really don't like about the umbrella enum pattern is that it forces people to write catch-all match arms. Of course there's no right or wrong here, just a matter of preference. I do see libraries choosing one over another all the time :)

PR #52, opened 2022-01-30
docs: add example for using `Account` to `README`

xJonathanLEI:
Resolves task 3 of #50:

> add example to `README` and mark feature as done

Note that this PR should be merged _after_ #51. Otherwise it doesn't make sense :))


milancermak (approved):
:shipit:

PR #54, opened 2022-02-02
feat: implement simple math operations for `UnsignedFieldElement`

xJonathanLEI:
This PR implements addition, subtraction, and multiplication between `UnsignedFieldElement`'s.

It makes it much easier to do something like "increasing nonce by one".

Division is left out on purpose for a future PR, as it needs to align with the somehow strange behavior of Cairo.


xJonathanLEI:
Wait there's no need to use `U512` in addition... It never overflows an `U256`.

milancermak (approved):
LGTM

Issue #53, opened 2022-01-31
DX improvements


milancermak:
While playing around with the lib, I came across a couple of peculiarities listed below. I think tackling them would result in a better developer experience. I'd like to know your opinion @xJonathanLEI, if you think they are worth dealing with.

#### 1) Print out a hash representation of `UnsignedFieldElement` (where it makes sense?):

Take the example from #52 of `dbg!(result);` This prints out something like this:
```
[src/main.rs:59] result = AddTransactionResult {
    code: TransactionReceived,
    transaction_hash: UnsignedFieldElement {
        inner: 2728449641502938754811783899429488071781078345184259994040934169808967186824,
    },
    address: None,
}
```
The issue is I can't copy-paste the transaction hash to [Voyager](https://goerli.voyager.online/) easily, I have to convert it to its hex representation first. Same applies to e.g. `block_hash` in various types. Is there a good solution for this?

#### 2) Build an abstraction over [`AcceptedOnL1` / `AcceptedOnL2`](https://github.com/xJonathanLEI/starknet-rs/blob/master/starknet-core/src/types/transaction_receipt.rs#L42-L43).

The idea is to prevent code like this:
```rust
  let tx = provider.get_transaction_status(tx_id).await.unwrap();

    if let TransactionStatus::AcceptedOnL2(b) = tx {
        dbg!(b.block_hash);
    } else if let TransactionStatus::AcceptedOnL1(b) = tx {
        dbg!(b.block_hash);
    }
```
since, normally, a TX gets accepted on L1 eventually, but I might not care when the transition happens. This might be potentially solved by resolving #9.

#### 3) Add `From`/`TryFrom` to streamline `UnsignedFieldElement` conversion
What do you think about implementing `From`/`TryFrom` traits to enable `UnsignedFieldElement::from("0xc0ffee")`, replacing `from_str` and `from_hex_str`? An added benefit is less need for `unwrap`ing. Similarly could be done to replace `try_from_bytes_be` though I don't think that's going to be used so often.

#### 4) Use `BlockId::Latest` instead of `None`
A bunch of provider functions take a `block_identifier: Option<BlockId>` argument. Setting it to `None` implies the latest block (however, only on the gateway level). This results in, subjectively, ugly code:
```rust
account.get_nonce(Some(BlockId::Number(56000))).await.unwrap();
account.get_nonce(None).await.unwrap();
```
[...]

PR #57, opened 2022-02-02
feat: explicit BlockId::Latest

milancermak:
This PR replaces the use of `Option<BlockId>` with an explicit `BlockId::Latest` variant, as discussed in point 4) in #53.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,10 @@ This workspace contains the following crates:
 ### Get the latest block from `alpha-goerli` testnet
 
 ```rust
-use starknet::{core::types::BlockId, providers::{Provider, SequencerGatewayProvider}};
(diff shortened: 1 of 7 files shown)


xJonathanLEI (changes requested):
The code changes look good 👍🏻

But the example in README won't compile after the change, as `BlockId` is not in scope.

milancermak:
Well spotted. Should be good now.

xJonathanLEI (changes requested):
It still doesn't compile lol. One extra bracket for the 3rd example.

xJonathanLEI:
And I fixed it for you :))

xJonathanLEI (approved):

PR #58, opened 2022-02-03
fix: deser or empty string to Option<UnsignedFieldElement>

milancermak:
An empty string `""` that's being deserialized with `hex_option` to a field of type `Option<UnsignedFieldElement>` currently deserializes to `Some(UnsignedFieldElement { inner: 0 })` instead of `None`. This PR fixes it.

I don't have an example of the API sending an empty string in e.g. a `block_hash`, I just found out when I was writing a test case by hand.

I wasn't sure how to unit test a custom deserializer, found [serde-test](https://docs.serde.rs/serde_test/) but I don't have a clue how to use it in this case. Then I found [this on SA](https://stackoverflow.com/questions/56985101/how-to-unit-test-a-deserialization-function-used-in-serdedeserialize-with) and adapted it, works pretty well IMO.

The matching code in hex_option's deserialize is kinda ugly though. Is there a prettier way how to do it?

### DIFF
--- a/starknet-core/src/serde/unsigned_field_element.rs
+++ b/starknet-core/src/serde/unsigned_field_element.rs
@@ -23,7 +23,7 @@ pub mod hex {
 }
 
 pub mod hex_option {
-    use serde::{Deserializer, Serializer};
+    use serde::{de::Error as DeError, Deserialize, Deserializer, Serializer};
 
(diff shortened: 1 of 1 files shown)


xJonathanLEI (approved):
Nice work

PR #60, opened 2022-02-04
refactor: tx types

milancermak:
This PR refactors transaction related types as outlined in #9 and also point 2) in #53.

The main changes are:
1) Replaces `TransactionStatus` with `BriefTransaction`
2) Replaces `TransactionWithStatus` with `FullTransaction`
3) Renames `Transaction` to `TransactionType`
4) Renames `TransactionStatusType` to `TransactionStatus`

I'm of course biased, but I think it results in types that are easier to comprehend and a lib that's simpler to use.


milancermak:
One thing I'm still not too happy about is the similar `DeployTransaction`/`InvokeFunctionTransaction` in `transaction.rs` and `transaction_request.rs`, but I'm not sure merging them is beneficial.

xJonathanLEI:
I'm pretty sure merging the tx request types will do more harm than good. You'll be essentially making it impossible to construct valid requests without consulting the Python code in `cairo-lang`, just to see what fields are needed for each type.

xJonathanLEI:
Sometimes cleaner looking code just isn't the easiest to use. The two variants are different enough to justify the separation IMO.

xJonathanLEI:
One possible way to merge them without hurting usability will be to use the builder pattern BTW.

starknet-core/src/types/transaction_request.rs
```diff
@@ -30,12 +23,19 @@ pub enum AddTransactionResultCode {
     TransactionReceived,
 }
 
+#[derive(Debug, Serialize)]
```
xJonathanLEI, line 26: Looks like all the changes to `transaction_request.rs` are just re-ordering things. Any specific reason for doing that?

xJonathanLEI (approved):
Looking good

starknet-core/src/types/transaction_request.rs
milancermak: Ah, I just reordered them alphabetically, when I was exploring the above mentioned types between `transaction.rs` and `transaction_request.rs`. Was easier to visually compare that way.

milancermak:
Yes, builder pattern FTW.

Issue #61, opened 2022-02-08
Unify field element types


xJonathanLEI:
Follow-up issue of [this comment](https://github.com/xJonathanLEI/starknet-rs/issues/55#issuecomment-1032829896).

We now have 2 field element types thanks to the work done in #13 and #33. There are a few reasons for unifying them:

- a cleaner API interface
- removal of the conversion layer when invoking `starknet-crypto` functions from `starknet-core`
- [additional benefits of adopting the `arkwork` ecosystem](https://github.com/xJonathanLEI/starknet-rs/issues/55#issuecomment-1032794772)

We should do this through a multi-step process:

- [x] replace `ff` with `ark-ff` in `starknet-crypto` to make sure crypto primitives are properly upheld
- [x] refactor the field element type from `starknet-crypto` into a new sub-crate `starknet-ff`
- [x] replace `UnsignedFieldElement` in `starknet-core` with the new type from `starknet-ff`

xJonathanLEI:
All tasks are resolved.

Issue #55, opened 2022-02-02
Replace `ff` with another library


xJonathanLEI:
Follow-up issue of #41. We can't publish anything to crates.io at the moment due to the `ff` issue.

Equilibrium opened [zkcrypto/ff#72](https://github.com/zkcrypto/ff/pull/72) and [zkcrypto/ff#73](https://github.com/zkcrypto/ff/pull/73), but unfortunately the repo seems to be inactive.

Given the situation, we should probably start experimenting with replacing `ff` with an alternative library. Doing so will introduce a breaking change to `starknet-crypto` but not any other crates.

milancermak:
I've noticed [this comment](https://github.com/zkcrypto/ff/issues/70#issuecomment-1027118408),  so there's hope for a new ff version, but only in a matter of weeks.

xJonathanLEI:
> High-level recommendation, making starknet-rs use [arkworks' traits](https://github.com/arkworks-rs/algebra/tree/master/ff) (cc @ZKProofs) would give us a lot of stuff for free in the future (e.g. STARK provers using the [LDT library](https://github.com/arkworks-rs/ldt))

Thanks for the advice! Currently `starknet-core` and `starknet-crypto` use different field element implementation: the `starknet-crypto` one is backed by `ff` and the `starknet-core` one is backed by `ethereum_types::U256`. I'm planning to unify them, and yes I'm exactly looking into `ark-ff` right now.

Basically I want to put this field element implementation into yet another sub-crate `starknet-ff`, which will be depended on by both `starknet-core` and `starknet-crypto`. `starknet-ff` will provide a field element type backed by `ark-ff`.

So yes we will adopt `ark-ff` and replace what's been done in #13.

xJonathanLEI:
Superseded by #61.

Issue #41, opened 2022-01-26
Release `starknet-crypto` - v0.1.0?


xJonathanLEI:
With #38, #39, and #40 merged, we've implemented all high-level crypto functions necessary for building a signer, which means the underlying `starknet-crypto` library is up to its job. Shall we release a `v0.1.0` to crates.io?

@milancermak What do you think?

milancermak:
DO IT 🤘

xJonathanLEI:
Hmmm turns out you cannot publish a crate to crates.io which depends on something that's not on crates.io (which makes sense).

We're using [eqlabs's fork of `ff`](https://github.com/eqlabs/ff) because the official `ff` crate clashes with `bitvec`.

milancermak:
Bummer. Do you want to work around it?

xJonathanLEI:
Yeah we gotta work something out lol. A few options off my head:

- ask eqlab to publish their fork to crates.io
- we fork from them and publish it to crates.io
- we try to figure out how to solve the `bitvec` clash (which seems unlikely to work)

xJonathanLEI:
I think I figured out option 3... Got something working locally using the official `ff` package.

xJonathanLEI:
I'm able to build and test the whole project by getting rid of `primitive-types` altogether, and using the underlying `uint` crate instead to construct my own `U256` (doesn't matter since it's an implementation detail anyways).

It works, but there's still an issue: this library will still clash with applications that use both this library and `primitive-types` (most of the widely used Parity crates depend on it directly or indirectly).

xJonathanLEI:
Probably a bad solution as I can see a lot of people doing exactly just that: using `ethers-rs` and `starknet-rs` side by side, for example.

milancermak:
Yeah, for sure, that's going to be a common combo.

xJonathanLEI:
Yep sounds good!

xJonathanLEI:
I think we should. Let's do it :)

One thing though: do you think we should align versions for all subcrates? For example, do we force bump all subcrates to 0.2.0 if we want to release 0.2.0 for `starknet`, our main crate?

I feel like doing this will make life easier for people trying to use the underlying crates directly, since they can simply use the same version for different crates.

But then on the other side, this means we will be creating new versions for some subcrates without any actual code changes, if they haven't been touched since the last main crate release. I feel like this is anti-pattern.

[...]

PR #63, opened 2022-02-15
feat: replace `ff` with `ark-ff`

xJonathanLEI:
Resolves task 1 of #61:

> replace `ff` with `ark-ff` in `starknet-crypto` to make sure crypto primitives are properly upheld

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,21 +3,70 @@
 version = 3
 
 [[package]]
-name = "addchain"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


xJonathanLEI:
Instead of using the `Fp` type directly, I'm using a wrapper type `FieldElement` such that we can have control over the public API interface.

Currently no `arkworks` traits are implemented on this wrapper type as suggested by [this comment](https://github.com/xJonathanLEI/starknet-rs/issues/55#issuecomment-1032794772), but it would be trivial to do so.

xJonathanLEI:
This will also unblock #41 since we no longer have non-crates-io dependencies.

milancermak (approved):
Big brain stuff this here, very nice work mate!

Issue #62, opened 2022-02-15
Add support for StarkNet v0.7.1


xJonathanLEI:
StarkNet `v0.7.1` is around the corner and I'm creating this issue to track support for this version.

TODOs:

- [x] support the new account contract
- [x] support new feeder gateway APIs `get_state_update` and `get_full_contract`

xJonathanLEI:
`v0.7.1` is now released to `alpha-goerli`. I'm starting to work on support for this version.

xJonathanLEI:
For task 2 I'm waiting on OpenZeppelin to update their [IAccount.cairo](https://github.com/OpenZeppelin/cairo-contracts/blob/main/contracts/IAccount.cairo) before proceeding.

xJonathanLEI:
For task 1 I still haven't figured out where the field has been added to LOL. @milancermak any idea?

milancermak:
The way I understood is it's *us* that should add a new `fee` field when creating a transaction. It's not something that's coming in the API response.

milancermak:
So probably a new field to `InvokeFunctionTransaction`, with a hardcoded value of 0 for now.

xJonathanLEI:
Hmmmm sounds.... reasonable. I mean I appreciate StarkWare providing the docs for preparation for the upcoming version, but it would be much nicer if it can be more detailed...

xJonathanLEI:
I just tried adding a `fee` field to `InvokeFunctionTransactionRequest` and setting it to `0` or `"0"` or `"0x0"`, and got:

```json
{
  "code": "StarkErrorCode.SCHEMA_VALIDATION_ERROR",
  "message": "Schema validation failed",
  "problems": {
    "fee": [
      "Unknown field."
    ]
  }
}
```

Looks like it's not how it works...

xJonathanLEI:
Got clarification from StarkWare that this is rather a change on the account contract. We will probably need to wait for OpenZeppelin to update their account contract first, otherwise we'll just be sending invalid signatures.

xJonathanLEI:
Looks like OpenZeppelin updated their [Account](https://github.com/OpenZeppelin/cairo-contracts/blob/e4b8baec2bb0f8c01f5a6bbc59b94d01efec3769/openzeppelin/account/Account.cairo) contract. I'll start working on task 1.

xJonathanLEI:
[...]

PR #65, opened 2022-02-16
refactor: adopt `serde_with`

xJonathanLEI:
This PR replaces `serde` modules with `serde_with` traits.

When working on #62, I encountered a type (`storage_diffs` from `get_state_update`) that uses `UnsignedFieldElement` as a `HashMap` key. To deserialize this type under the existing `serde` solution, I'll have to build yet another module that targets `HashMap<UnsignedFieldElement, _>` specifically, similar to what's been done with `Vec<UnsignedFieldElement>`. That's really annoying and makes code very ugly. So I've decided to use the magic from `serde_with` to solve it for good.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -386,6 +386,41 @@ dependencies = [
 ]
 
 [[package]]
+name = "darling"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


xJonathanLEI:
Oops I was basing on the old `master`. Rebasing.

milancermak (approved):
LGTM

PR #66, opened 2022-02-16
feat: support new APIs from StarkNet `v0.7.1`

xJonathanLEI:
Resolves task 2 of #62:

> support new feeder gateway APIs `get_state_update` and `get_full_contract`

I had to move `starknet_contract::ContractArtifact` to `starknet_core::types::ContractArtifact` since `get_full_contract` returns it. Did a small refactor on that. Also moved `test-data` from `starknet-contract` to `starknet-core` and turned the `starknet-contract` one into a symlink.


milancermak:
I tried fetching `http://alpha4.starknet.io/feeder_gateway/get_state_update` without specifying a block ID. It returned a JSON without a `block_hash` element which breaks the deserialization of `StateUpdate`.

xJonathanLEI:
You're right. Thanks for spotting.

xJonathanLEI:
Though I have to say this is extremely strange. I tried querying with `blockNumber=pending` and without any query param and they return different things (different `old_root` and yes I did retry and still the same, so it's not a pending block getting finalized by chance), which means querying without param does return the latest block as expected. The strange part is the latest block, despite being finalized on L2, does not come with a block hash.

xJonathanLEI:
And even stranger, I tried querying the latest block with `blockNumber=xxxx` where `xxxx` comes from `get_block`, and without query param. They returned the exact same data *except* `block_hash`. This is weird...

xJonathanLEI:
In any case, the issue is now resolves :)

milancermak:
Yeah, the API is whack 🤷‍♂️

milancermak (approved):
Looking good 👏🏼

PR #69, opened 2022-02-21
refactor: replace `UnsignedFieldElement` with `starknet-ff`

xJonathanLEI:
Resolves task 3 of #61:

> replace `UnsignedFieldElement` in `starknet-core` with the new type from `starknet-ff`

This PR:

- ports all relevant functionalities on `UnsignedFieldElement` to `starknet_ff::FieldElement`
- removes `UnsignedFieldElement` and changes to use `starknet_ff::FieldElement` instead


xJonathanLEI:
Sorry for changing a ton of files as this type is literally used everywhere LOL.

xJonathanLEI:
I decided to make the `Debug` view of `FieldElement` a hex representation, which kinda helps with item 1 on #53? But it also means now you need to specifically write code for it to get the decimal representation LOL.

milancermak:
Sweet! Will have a look tomorrow.

starknet-ff/src/lib.rs
```diff
@@ -1,56 +1,166 @@
...
     ///
-    /// * `bytes`: The byte array in **big endian** format
-    pub fn from_bytes_be(bytes: &[u8; 32]) -> Option<Self> {
-        let mut bits = [false; 32 * 8];
-        for (ind_byte, byte) in bytes.iter().enumerate() {
-            for ind_bit in 0..8 {
-                bits[ind_byte * 8 + ind_bit] = (byte >> (7 - ind_bit)) & 1 == 1;
-            }
+    /// * `bytes`: The byte slice in **big endian** format
+    pub fn from_slice_be(bytes: &[u8]) -> Result<Self, FromByteSliceError> {
```
milancermak, line 142: Do you think it's necessary to have this function? In the lib itself it's used only once. I can't judge how useful it is for the lib uses, but maybe just having `from_bytes_be` is ok?

milancermak (approved):
*chef's kiss*

starknet-ff/src/lib.rs
xJonathanLEI: I think we should keep the public API surface as small as possible as removing functions is a breaking change. Lemme remove this for now.

Before this I could have just made it `pub(crate)`, but now since `FieldElement` is a standalone crate, I have to start thinking carefully about the interface I guess.

xJonathanLEI: Done. Replaced the occurrence with an `unsafe` block instead.

PR #71, opened 2022-02-23
doc: fix broken `IAccount` link on README

xJonathanLEI:
OpenZeppelin restructured their repo so the `IAccount` link no longer works. I don't really want to use a permalink here as `IAccount` can evolve over time.


milancermak (approved):
👍

PR #72, opened 2022-02-24
refactor: updating provider types for Cairo 0.7.1

### LINKED ISSUE #64: Match response object type names with `cairo-lang`
With the release of StarkNet `v0.7.1`, I noticed that a new file [response_objects.py](https://github.com/starkware-libs/cairo-lang/blob/ed6cf8d6cec50a6ad95fa36d1eb4a7f48538019e/src/starkware/starknet/services/api/feeder_gateway/response_objects.py) has been added to [cairo-lang](https://github.com/starkware-libs/cairo-lang) which contains gateway response object type definitions. While they're still using `JsonObject` for most requests, some of them are now typed and thus have official names.

We should probably update our code to match with their names, for the sake of making it easier for people familiar with [cairo-lang](https://github.com/starkware-libs/cairo-lang) to start using our library.

milancermak:
Resolves #64.

I've made the following changes to better match our data types to the ones from feeder gateway:

* Added a `status` field to the `Block` type; `Block` now matches the `StarknetBlock` in cairo-lang.
* Updated `FullTransaction` - it's been renamed to `TransactionInfo` to match cairo-lang nomenclature. I've also added a new `transaction_failure_reason` field. I've also renamed its field `transaction` to `r#type`, see discussion point 2 below.
* Renamed `BriefTransaction` to `TransactionStatusInfo` to be aligned with the above, even though the response type for `get_transaction_status` in cairo-lang is still a `JsonObject`. I've also added the `transaction_failure_reason` to this type (with a test case). I kept the field name same as in `TransactionInfo` for consistency reason, even though in the raw type it's named `tx_failure_reason` 🤔 
* Added a `Constructor` case to the `EntryPointType` enum since it's also present in [cairo-lang](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/services/api/contract_definition.py#L25) but to be honest I have no idea how to actually get it in a response. I only found it by chance and added it for completeness sake (it's been there since cairo 0.5.0).
* Updated `Receipt` with 2 new fields - `transaction_failure_reason` (same as above) and a new `l1_to_l2_consumed_message` that's present in cairo-lang, but once again, I didn't find a way how to replicate it.
* I've added inline comments to `BlockStatus` and `TransactionStatus` enums, taken from cairo-lang repo.

Discussion points:

I'd like to hear your opinion on the following 2 things @xJonathanLEI 

1) Should we keep `ConfirmedReceipt`? It's a nicer-to-work-with subset of the `Receipt` type we introduced. cairo-lang doesn't have anything like that, they only use a single `TransactionReceipt` object, so we could do the same. From a lib maintainer perspective, having it around is annoying because relevant changes to `Receipt` should be replicated to `ConfirmedReceipt` as well and it's easy to forget (FWIW, I checked a lot of responses and none contained the new fields from `Receipt` so I didn't add these new fields to `ConfirmedReceipt`). Plus you raised a good point the other day of keeping the public API surface small. However from a user perspective, it's more pleasant to work with less optional fields. So IDK, I'm 50/50 on this one :)

2) As mentioned above, I renamed `transaction` to `r#type` in the new `TransactionInfo` type. In the raw JSON, it's called `transaction`, but its structure depends on the `type` and its of `TransactionType` so I found it appropriate. In cairo-lang, they also use `internal` or `specific` in various places to store the deploy / invoke fields. If you have different suggestion on how to name this, I'm all ears 👂


xJonathanLEI:
Dood... The guys at StarkWare responsible for the gateway API really have issue getting stuff consistent... I'm mad :(

milancermak:
Yeah, one would think consistency wouldn't be a problem in a blockchain company :D

xJonathanLEI (changes requested):
Code is all good! Thanks for the hard work!

One very minor thing about style though: from what I see, the convention in Rust seems to be starting comments with a capital letter (sentence case), even if the comment itself is not a complete sentence. I think we should probably follow :)

milancermak:
Sure thing

xJonathanLEI (approved):
All good! 🚢

PR #75, opened 2022-02-24
fix: incorrect `FieldElement` decimal formatting

### LINKED ISSUE #74: Bug: incorrect JSON serialization on `FieldElement`
So when working on #62 I found out about a critical bug on the `FieldElement` type that's basically breaking all contract deployments and calls. This code doesn't work:

```rs
let num = FieldElement::from_hex_be("0x1").unwrap();
let text = serde_json::to_string(&num).unwrap();
assert_eq!(text, "\"1\"");
```

Fails with:

```log
thread 'tests::test_json_ser' panicked at 'assertion failed: `(left == right)`
  left: `"\"0\""`,
 right: `"\"1\""`', starknet-ff/src/lib.rs:450:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

xJonathanLEI:
Resolves #74.

This PR fixes the critical bug of incorrect decimal representation of `FieldElement`.

### DIFF
--- a/starknet-ff/src/lib.rs
+++ b/starknet-ff/src/lib.rs
@@ -242,7 +242,11 @@ impl Display for FieldElement {
         let ten = U256::from_u8(10u8);
 
         loop {
-            let digit = (current.checked_rem(&ten)).unwrap().to_uint_array()[0] as u8;
+            let digit = if current < ten {
+                current.to_uint_array()[0] as u8
+            } else {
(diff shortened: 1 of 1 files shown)


xJonathanLEI:
@milancermak Sorry buddy I'm forcing this into `master` considering it's a critical bug (people can be deploying account contracts with incorrect public keys).

milancermak:
No worries. Good call and thanks for the fast fix ⚡

Issue #76, opened 2022-03-01
Cannot deploy the new OpenZeppelin account contract


xJonathanLEI:
Tried using `starknet-rs` to deploy the latest [account contract from OpenZeppelin](https://github.com/OpenZeppelin/cairo-contracts/blob/5731e7134e755ad9b821963bf7d20f06acd1d355/openzeppelin/account/Account.cairo).

The transaction is [not working](https://alpha4.starknet.io/feeder_gateway/get_transaction?transactionHash=0x0232db7959b3ccc22b6e1321fc412025a7d9733b90db809c29e47601b5ef5e3f):

```json
{
  "status": "REJECTED",
  "transaction_failure_reason": {
    "tx_id": 1256114,
    "code": "TRANSACTION_FAILED",
    "error_message": "Error at pc=0:45:\nGot an exception while executing a hint.\nCairo traceback (most recent call last):\nUnknown location (pc=0:509)\nUnknown location (pc=0:499)\nUnknown location (pc=0:280)\nUnknown location (pc=0:242)\nUnknown location (pc=0:219)\n\nTraceback (most recent call last):\n  File \"<hint6>\", line 5, in <module>\nAssertionError: normalize_address() cannot be used with the current constants."
  },
  "transaction": {
    "contract_address": "0x615161661cdbea1c112383ce703d33299bef0e796903b88ec5601010ee22091",
    "contract_address_salt": "0x6c8035ea0cc6d4b0d933969db1367d7380adcaa462184b814f2def2bd9adf5",
    "constructor_calldata": [
      "1"
    ],
    "transaction_hash": "0x232db7959b3ccc22b6e1321fc412025a7d9733b90db809c29e47601b5ef5e3f",
    "type": "DEPLOY"
  }
}
```

xJonathanLEI:
Did some testing. I started out trying to deploy the artifact generated by the `generate_artifacts.sh` script with the official StarkNet CLI. It fails with the same error.

Then I tried directly compiling the account contract with `starknet-compile` **and** deploying with the official CLI. Much to my surprise, it works.

After some more testing, it turns out piping the artifact through `jq` is what causes the issue... The `generate_artifacts.sh` script uses `jq` to turn each artifact into a single line (so that Git diffs for example aren't a complete mess).

I tried using `prettier` to format the artifact instead, and it can still be successfully deployed with the official CLI.

I'm still in the process of trying to figure out what exactly has `jq` broken.

[...]

PR #80, opened 2022-03-06
fix: corrupted contract artifacts

xJonathanLEI:
This PR fixes the scientific notation issue of `jq` mentioned in https://github.com/xJonathanLEI/starknet-rs/issues/76#issuecomment-1058153538 by... not using `jq` at all.


milancermak (approved):
LGTM

PR #82, opened 2022-03-08
feat: support new account contract with StarkNet 0.7.1

xJonathanLEI:
Resolves task 1 of #62:

> support the new account contract

This PR:

- adds support for the [new account contract](https://github.com/OpenZeppelin/cairo-contracts/blob/2ebdc7352471c53838c1c0d39f3ab2da802b5881/openzeppelin/account/Account.cairo), which **supports multi-call**; and
- changes to use Git submodule for OpenZeppelin's contracts since the account contract is no longer a single-file contract; and
- updates `cairo-lang` to `0.7.1` for `generate_artifacts.sh`.


milancermak:
Sorry this took so long, was trying to understand how it all works.

It looks good overall, but I have one question. Shouldn't `execute` be done via `call_contract` and return a `Vec<FieldElement>` since [`__execute__`](https://github.com/OpenZeppelin/cairo-contracts/blob/2ebdc7352471c53838c1c0d39f3ab2da802b5881/openzeppelin/account/Account.cairo#L112) returns an array of felts?

xJonathanLEI:
Nah `execute` is meant to dispatch a transaction, which uses `add_transaction` instead of `call_contract`. `call_contract` is `eth_call`, which only simulates the transaction without adding it to the network.

You're talking about the contract function signature, but just like in L1, there's no way to get contract return values from a transaction; the only way to do that would be to call the function on-chain, from within a transaction.

xJonathanLEI:
LOL you sure are :)

I guess one never gets enough sleep in crypto.

PR #83, opened 2022-03-09
ci: add CI pipeline for running WASM tests

xJonathanLEI:
This PR:

- makes all `#[test]` tests runnable for `wasm32-unknown-unknown` through `wasm-bindgen-test`; and
- adds yet another CI pipeline for running these tests via `wasm-pack test`.

We have added wasm-specific code for the WebAssembly support, so it's important we keep testing stuff under wasm too.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -65,3 +65,48 @@ jobs:
         with:
           command: test
           args: --all --target ${{ matrix.target }}
+
+  wasm-test:
+    name: WASM tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout source code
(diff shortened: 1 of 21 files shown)


milancermak (approved):
Sweeeeeet!

What's the rule for `#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]`? When should it be used and when not?

xJonathanLEI:
I have to add `wasm_bindgen_test` to make the tests work for `wasm-pack` because unfortunately it doesn't work with `#[test]`.

At the same time I also don't want this to impact compilation time for non-wasm test runs, so I made it conditional :)

Issue #85, opened 2022-03-10
Add support for StarkNet v0.8.0


xJonathanLEI:
OK so as we're wrapping up our work on supporting #62, StarkNet v0.8.0 is already around the corner... Creating this issue to track the progress on supporting v0.8.0.

Can barely keep up :(

TODOs:

- [x] resolve breaking changes
- [x] add support for new gateway APIs
- [x] add support for paying transaction fees

xJonathanLEI:
Looks like the StarkWare folks made quite some breaking changes. The ones I've discovered so far reading the code diff:

- some fields are now hex-encoded: `calldata`, `constructor_calldata` and more;
- the `tx_id` concept is gone. You can no longer query transactions or their receipts etc. by transaction ID; funny enough, `get_transaction_hash_by_id` and `get_transaction_id_by_hash` still work;
- as a result, the field`tx_id` is removed from `TransactionFailureReason`.

Most pre-0.8.0 clients should have been broken by this update. We probably need more tests.

xJonathanLEI:
The library now fully supports StarkNet v0.8.0

PR #87, opened 2022-03-16
dev: resolves breaking changes of StarkNet v0.8.0

xJonathanLEI:
Resolves task 1 of #85:

> resolve breaking changes

This PR:

- re-generates raw gateway responses;
  - block for `get_block/2_with_messages.txt` was specifically changed to not use block `39232` because there seems to be a bug with that block: a transaction with message is now shown without any;
- makes all fields in `BuiltinInstanceCounter` optional since I observed that all the fields are gone when `n_steps` is zero;
- changes some fields to use hex-encoding to comply with StarkNet 0.8.0;
- removes ability to query transaction data with transaction ID, since it's removed from StarkNet 0.8.0.


xJonathanLEI:
Confirmed with StarkWare that the missing message thing is indeed a bug on their side.

milancermak (approved):
LGTM

PR #92, opened 2022-03-24
feat(starknet-providers): Make Provider Futures Send

abigger87:
Changes the `#[async_trait(?Send)]` attribute to `#[async_trait]`

### DIFF
--- a/starknet-accounts/src/account.rs
+++ b/starknet-accounts/src/account.rs
@@ -4,7 +4,7 @@ use async_trait::async_trait;
 use starknet_core::types::{AddTransactionResult, BlockId, FieldElement};
 use std::error::Error;
 
-#[async_trait(?Send)]
+#[async_trait]
 pub trait Account {
     type GetNonceError: Error + Send;
     type ExecuteError: Error + Send;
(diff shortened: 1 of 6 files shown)


xJonathanLEI:
Hi thanks for the PR. Could you explain the rationale for this change? I believe it was added to make WASM work.

Also it looks like our there's something wrong with our GitHub Actions settings. CI seems to not be running.

xJonathanLEI:
Maybe related?

https://www.githubstatus.com/incidents/y5hdmv0p49x3

Let me try closing and re-opening the PR to see of it can be triggered successfully. Sorry for the notification spam.

xJonathanLEI:
OK it _is_ related.

abigger87:
No worries - the reason I wanted to open this is to be able to send the provider futures across runtime threads!

xJonathanLEI:
It looks like our current CI pipeline doesn't attempt to build the whole project for WASM... it's not catching the build failure as a result.

I just tried running `cargo build --target wasm32-unknown-unknown` on your branch and it fails.

xJonathanLEI:
I just updated our CI workflow to include building for wasm, but I think this PR is still running the old version. That's why it's all green. Actually your branch fails the wasm build.

Might need to use conditional attribute instead?

PR #97, opened 2022-03-26
feat(starknet-contracts): Send + Sync Safe Random Number Generation

abigger87:
This PR introduces `Send + Sync` safe random number generation by using the [rug](https://docs.rs/rug/latest/rug/index.html) crate.


xJonathanLEI:
Hey I'm pretty sure you added a Linux-only crate. It's our goal to make `starknet-rs` work cross-platform (and that means `macos` + `linux` + `windows` + `wasm32` atm). I think CI is failing because of this.

PR #100, opened 2022-04-05
doc: mention L2 faucet to pay for TX fees

milancermak:
Adding a line about the Goerli faucet and TX fees to the example in README


xJonathanLEI (approved):
👍🏻

PR #104, opened 2022-04-11
feat: add fee estimate buffer

### LINKED ISSUE #103: Add buffer to `max_fee` from fee estimate
Quoting StarkWare:

> Hey guys! quick note about fees following some trouble we had with rejected transactions.
>
> Currently the fee charged is determined by the marginal L1 cost (proof + on chain data), and is thus measured in gas. The 
`estimate_fee` endpoint returns the price in wei, using the current gas price. 
>
> If no buffer is used between the result of estimate_fee and `max_fee` used for the transaction, then a negligible change in gas prices can cause the transaction to be rejected, as the estimated fee at the time of sequencer execution will be higher than max_fee. We suggest, in the different libraries, to set `max_fee=1.1*estimated_fee`. Nothing special about 10% (good magic number), the difference depends on the average time a transaction lives in the "mempool", and the expected change in gas price during this interval.
>
> We're also considering adding some leeway in the sequencer itself in 0.8.2, i.e. having him accept transactions with |max_fee-estimated_fee| < epsilon (can be multiplicative).

xJonathanLEI:
Resolves #103.

### DIFF
--- a/starknet-accounts/src/single_owner.rs
+++ b/starknet-accounts/src/single_owner.rs
@@ -180,8 +180,9 @@ where
             .await
             .map_err(ExecuteError::ProviderError)?;
 
+        // Adds 10% fee buffer
         let add_transaction_request = self
-            .generate_invoke_request(calls, nonce, fee_estimate.amount.into())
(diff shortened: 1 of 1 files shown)


milancermak (approved):
👍

PR #105, opened 2022-04-11
feat: resolve StarkNet v0.8.1 API changes

### LINKED ISSUE #99: Add support for StarkNet v0.8.1
StarkNet 0.8.1 is about to release. Creating this issue to track integration progress.

See comments below. It looks like we only need to resolve API changes for this version.

xJonathanLEI:
Resolves #99.


starknet-core/src/types/transaction_receipt.rs
```diff
@@ -24,6 +24,9 @@ pub struct Receipt {
     #[serde_as(as = "UfeHex")]
     pub transaction_hash: FieldElement,
     pub transaction_index: Option<u64>,
+    #[serde(default)]
+    #[serde_as(as = "Option<UfeHex>")]
+    pub actual_fee: Option<FieldElement>,
```
milancermak, line 29: The `actual_fee` field also appears in a [block](https://alpha4.starknet.io/feeder_gateway/get_block?blockId=PENDING) under `transaction_receipts`, but it's missing in our `ConfirmedReceipt` object.

starknet-core/src/types/transaction.rs
```diff
@@ -64,6 +64,10 @@ pub struct DeployTransaction {
     pub contract_address: FieldElement,
     #[serde_as(as = "UfeHex")]
     pub contract_address_salt: FieldElement,
+    // Field marked optional as `get_block` doesn't include it yet. Drop optional once resolved.
```
milancermak, line 67: Yeah, so I guess this will stay optional 🥶

milancermak (approved):
Other than the missing field all looks good for 🚢

starknet-core/src/types/transaction_receipt.rs
xJonathanLEI: I actually checked before excluding... I think it's the same case as `class_hash`. Only new blocks have it I guess.

Let me double check and create separate test data for these stuff.

xJonathanLEI: I added assertions that old blocks don't have these fields so that we won't miss them fixing it, if ever.

PR #107, opened 2022-04-14
ci: run clippy checks on all targets

xJonathanLEI:
I just realized our CI has not been checking non-default targets like tests and benches. This PR fixes it along with some clippy warnings discovered.


milancermak (approved):
Huh, good catch. I thought we're already doing this.

PR #109, opened 2022-04-17
feat: add util function for converting Cairo string to felt

xJonathanLEI:
This PR adds a simple utility function for converting [Cairo short strings](https://www.cairo-lang.org/docs/how_cairo_works/consts.html#short-string-literals) to `FieldElement`.


milancermak (approved):
Could you add test cases for the errors as well? 

Otherwise looking good as always.

xJonathanLEI:
Error cases added.

PR #110, opened 2022-04-17
feat: use transaction_hash for account signatures

### LINKED ISSUE #102: Update for new account signature scheme
I just realized that OpenZeppelin [changed how signatures are validated](https://github.com/OpenZeppelin/cairo-contracts/commit/574bc03474ff6deba186dbce3e17ffa62f7d7286) a while ago. I wonder if there is even a standard for account signatures, as it looks like Argent X is [still using the old hashing](https://github.com/argentlabs/argent-contracts-starknet/blob/e0ae18af621a41b9cfa690df89c625b848693b06/contracts/ArgentAccount.cairo).

Not sure what's the appropriate action on our side. If there's a standard, we should probably stick to the standard. Otherwise we will have to build different `Account` implementations to cater for each major implementation.

Creating this issue to track progress. Will have to gather more info before concrete actions.

xJonathanLEI:
Resolves #102.

This PR:

- regenerates contract artifacts with updated [cairo-contracts](https://github.com/OpenZeppelin/cairo-contracts) and `starknet-compile` v0.8.1; and
- changes to use a locally computed transaction hash for account signatures.


starknet-accounts/src/single_owner.rs
```diff
@@ -116,6 +93,21 @@ where
         }
         execute_calldata.push(nonce); // nonce
 
+        let transaction_hash = compute_hash_on_elements(&[
+            PREFIX_INVOKE,
+            FieldElement::ZERO, // version
+            self.address,
+            get_selector_from_name("__execute__").unwrap(),
```
milancermak, line 100: Wouldn't it make sense to have this as a const as well? Could be reused just below, on line 113.

milancermak (approved):
Nice

starknet-accounts/src/single_owner.rs
xJonathanLEI: Makes sense. Even better, I would like to make these util functions themselves const too. Will look into that later.

PR #111, opened 2022-04-18
Dev/deny unknown fields

### LINKED ISSUE #73: Add `deny_unknown_fields` to test code
Following up on https://github.com/xJonathanLEI/starknet-rs/pull/72#issuecomment-1049971949.

This will allow us to ensure out data types are not missing fields, while not making user code broken by new fields introduced by StarkNet.

xJonathanLEI:
Resolves #73.

The attribute is not added to `AbiEntry` related types (e.g. `Constructor`) because those types are affected by the `serde` bug. Deserialization is broken if the attribute is applied.

A missing field has been discovered in the process: `l1_to_l2_consumed_message` is missing from `ConfirmedReceipt`. It's been fixed.


milancermak (approved):
Very elegant solution this.

PR #113, opened 2022-04-22
feat: use builder pattern for `Account`

### LINKED ISSUE #101: Use builder pattern for Account transactions
Follow up issue on #98. Instead of hard-coding a fee estimation process, we should adapt the builder pattern to allow users to specify transaction fees.

Such a pattern should also be applied to nonce fetching. Currently `execute()` forces callers to supply a nonce, meaning it's the caller's responsibility to call `get_nonce()` before calling `execute()`. It makes the `Account` trait hard to use.

xJonathanLEI:
Resolves #101.

Now you can just do `account.execute(&[Call{...}]).send().await`, and inject `nonce` or `max_fee` if you want.

### DIFF
--- a/examples/mint_tokens.rs
+++ b/examples/mint_tokens.rs
@@ -1,10 +1,6 @@
 use starknet::{
     accounts::{Account, Call, SingleOwnerAccount},
-    core::{
-        chain_id,
-        types::{BlockId, FieldElement},
-        utils::get_selector_from_name,
-    },
+    core::{chain_id, types::FieldElement, utils::get_selector_from_name},
     providers::SequencerGatewayProvider,
     signers::{LocalWallet, SigningKey},
 };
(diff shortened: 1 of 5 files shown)


milancermak (approved):
Very cool overall. I'm not quite fond of the name of `execute` now as it evokes things actually happening, so I wouldn't expect the need to still do `send` afterwards. Maybe `prepare` would be more appropriate, but I don't wanna bikeshed.

One thing that came to my mind when I was playing around with the code is if we should provide a simpler way how to estimate the fee of the calls - basically the same thing you do when `get_max_fee` is None, just wrapped in a single function. I don't have a use case for it though 🤷‍♂️

xJonathanLEI:
For the fee estimation part, I think we can add `estimate_fee()` alongside `send()` on `AttachedAccountCall`, very similar to what's done in `ethers-rs`.

PR #119, opened 2022-05-11
feat(providers): Instantiate localhost provider

exp-table:
Added another function to create an instance of `SequencerGatewayProvider` with the appropriate args for a local dev node.

I am referring to the default parameters nile is [using](https://github.com/OpenZeppelin/nile/blob/main/src/nile/core/node.py).

### DIFF
--- a/starknet-providers/src/sequencer_gateway.rs
+++ b/starknet-providers/src/sequencer_gateway.rs
@@ -58,7 +58,7 @@ impl SequencerGatewayProvider {
         )
     }
 
-    pub fn starknet_alpha_localhost() -> Self {
+    pub fn starknet_nile_localhost() -> Self {
         Self::new(
             Url::parse("http://127.0.0.1:5000/gateway").unwrap(),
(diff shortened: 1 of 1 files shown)


starknet-providers/src/sequencer_gateway.rs
```diff
@@ -57,6 +57,13 @@ impl SequencerGatewayProvider {
             Url::parse("https://alpha4.starknet.io/feeder_gateway").unwrap(),
         )
     }
+
+    pub fn starknet_alpha_localhost() -> Self {
```
xJonathanLEI, line 61: Can you rename the function to `starknet_nile_localhost`? I feel like it's more descriptive that way, as there are other local networks like the devnet.

xJonathanLEI (changes requested):

starknet-providers/src/sequencer_gateway.rs
exp-table: Good point! Pushed the change.

xJonathanLEI (approved):
LGTM

Issue #35, opened 2022-01-21
Port `pathfinder`'s Pedersen hash performance improvements


xJonathanLEI:
The Pedersen hash implementation added by #33 was refactored from [b091cb8 @ `pathfinder`](https://github.com/eqlabs/pathfinder/tree/b091cb889e624897dbb0cbec3c1df9a9e411eb1e). [Changes](https://github.com/eqlabs/pathfinder/compare/b091cb889e624897dbb0cbec3c1df9a9e411eb1e...3331f26598640822838e1a49857a8492038c5620) have been made to their implementation that they claim to offer a 10x performance improvement.

xJonathanLEI:
Just ran their benchmark on my machine.

Before ([b091cb8](https://github.com/eqlabs/pathfinder/tree/b091cb889e624897dbb0cbec3c1df9a9e411eb1e)):
```log
pedersen_hash           time:   [3.1228 ms 3.1455 ms 3.1732 ms]
```

After ([3331f26](https://github.com/eqlabs/pathfinder/tree/3331f26598640822838e1a49857a8492038c5620)):
```log
pedersen_hash           time:   [239.06 us 240.10 us 241.21 us]
```

PR #126, opened 2022-05-29
Formatting and y-coordinate

rhobro:
I have implemented fmt::Display for Signatures so that they can be easily represented by strings.

I have written the method on EcPoint to calculate the y coordinate of the stark point.

Exposed Signature


xJonathanLEI:
Hi @rhobro thanks for the PR.

While I agree that `Signature` should indeed be exposed (since it's actually already part of the public API with `sign`), `EcPoint` is currently treated as an implementation detail that I do not want to expose so that we can change the curve implementation without breaking users.

In fact, there's one work in progress (#35) that will change exactly that for performance reasons.

rhobro:
Ok that's fair. Is it possible in pr to accept some changes but not others? Hahaha it is my first one.

How will #35 allow calculating the y coordinate?

xJonathanLEI:
It's not possible. You have to update your PR accordingly.

xJonathanLEI:
As for `y` calculation, it's actually part of `EcPoint::from_x()`, just not as a standalone function.

xJonathanLEI:
If you _really_ need to use `EcPoint` in your code, you'll have to fork the lib unfortunately.

rhobro:
I'll keep that in mind. Please do ping me if you expose a function to calculate y in the future.

I'll update my PR to include only the agreed upon changes.

Out of interest, what is the difference between sign and ecdsa_sign? I believe they are in starknet-core.

xJonathanLEI:
`sign` is _NOT_ in `starknet-core`. It's only in `starknet-crypto` that's not re-exported through `starknet-core`. It's not exported because it's a low-level crypto primitive that's dangerous. This is also noted in the `starknet-crypto` README.

`ecdsa_sign` is indeed exposed. It uses RFC 6979 to generate a deterministic scalar `k` before calling `sign`.

PR #152, opened 2022-06-08
feat: export BlockStatus

fracek:
Block status is also needed when interacting with the new json rpc.﻿


xJonathanLEI:
Thanks for the PR. BTW JSON-RPC support is being worked on on my side too :)

xJonathanLEI (approved):
LGTM

fracek:
> Thanks for the PR. BTW JSON-RPC support is being worked on on my side too :)

No worries, I just need two methods so I'm hacking something quick and dirty.

PR #164, opened 2022-06-15
feat: add macros crate with `selector` macro

xJonathanLEI:
This PR adds a proc macro `selector` so that:

- users no longer need to call `unwrap()` on `get_selector_from_name` for string literals; and
- selector values can now be `const`.


xJonathanLEI:
Debugged locally and the `starknet_call` failure was a 502. Changed to a local node and it passes. So the CI failure is irrelevant.

PR #186, opened 2022-06-28
fix(sequencer_gateway): derive Clone for sequencer SequencerGatewayPr…

tdelabro:
```rust
pub struct SequencerGatewayProvider {
    client: Client,
    gateway_url: Url,
    feeder_gateway_url: Url,
}
```
`Client` and `Url` are both Clone. So it can be derived.
It is used as provider in `SingleOwnerAccount`, which is also Clone.

### DIFF
--- a/starknet-providers/src/sequencer_gateway.rs
+++ b/starknet-providers/src/sequencer_gateway.rs
@@ -17,6 +17,7 @@ use starknet_core::{
 use thiserror::Error;
 use url::Url;
 
+#[derive(Clone)]
 pub struct SequencerGatewayProvider {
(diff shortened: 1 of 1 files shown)


xJonathanLEI:
It _can_ be `Clone` but I generally don't recommend doing that, as you would lose the ability to share the HTTP connection for faster requests.

Allowing `Clone` is good for people who know what they're doing, but generally you should just do wrap it inside an `Arc`, since `Arc<SequencerGatewayProvider>` implements `Provider` too, so you can use it inside `SIngleOwnerAccount`, for example.

PR #187, opened 2022-06-28
Expose provider in account trait

tdelabro:
The `Account` trait entirely on having a provider and a signer.
But atm we cannot get the provider on the signer out of the struct that implements `Account`. It is not handy at all.
This fix it

### DIFF
--- a/starknet-accounts/src/single_owner.rs
+++ b/starknet-accounts/src/single_owner.rs
@@ -35,7 +35,7 @@ const SELECTOR_EXECUTE: FieldElement = FieldElement::from_mont([
 pub struct SingleOwnerAccount<P, S>
 where
     P: Provider + Send,
-    S: Signer + Send + Clone,
+    S: Signer + Send,
 {
     provider: P,
     #[allow(unused)]


xJonathanLEI (commented):
Thanks for the PR but:

1. While I agree it's handy to be able to grab `provider` and `signer` out of a `SingleOwnerAccount`, I don't believe the same is true for the `Account` trait in general. You could in theory implement an account, say, an offline multi-sig account, where none of these are applicable. The `Account` trait was created to mirror the structure of the standard `IAccount` interface only.
2. For the implementation on the `SingleOwnerAccount` side, I don't like that:
  a. it adds the `Clone` requirement on `P` and `S`; and
  b. it forces the caller to do a potentially expensive clone.
  Callers should only grab a reference to the provider/signer and decide to clone it or not.

xJonathanLEI:
I'm not convinced that the `Account` trait should contain methods for getting provider or signer, but it should be totally fine to have `WithProvider` and `WithSigner` traits separately from `Account` itself.

PR #204, opened 2022-09-01
Deserialize l1 to l2 message without nonce

fracek:
Tests were missing a case in which the `L1ToL2Message`'s `nonce` field was missing. This happens, for example, [at block 1564)(https://alpha4.starknet.io/feeder_gateway/get_block?blockNumber=1564).


fracek:
Ok done!

PR #206, opened 2022-09-05
Implement changes for starknet v0.10

fracek:
I added the new fields as optional so that the library is still compatible with mainnet. Notice the PR doesn't include transaction v1 yet, I just wanted to start sharing the work on this.

Resolves task 1 of #207:

> resolve gateway API changes


xJonathanLEI:
That's fast. I was just about to start working on this. Thanks!

starknet-core/src/types/transaction.rs
```diff
@@ -114,6 +118,31 @@ pub struct InvokeFunctionTransaction {
     pub transaction_hash: FieldElement,
     #[serde_as(as = "UfeHex")]
     pub max_fee: FieldElement,
+    #[serde(default)]
+    #[serde_as(as = "Option<UfeHex>")]
+    pub nonce: Option<FieldElement>,
+    #[serde(default)]
+    #[serde_as(as = "Option<UfeHex>")]
+    pub version: Option<FieldElement>,
```
xJonathanLEI, line 126: Does the `version` field really have to be optional? Looks like it has been retrospectively applied to older blocks as well. Will check to make sure though.

starknet-core/src/types/state_update.rs
```diff
@@ -27,6 +27,9 @@ pub struct StateDiff {
     pub deployed_contracts: Vec<DeployedContract>,
     #[serde_as(as = "Vec<UfeHex>")]
     pub declared_contracts: Vec<FieldElement>,
+    #[serde(default)]
+    #[serde_as(as = "HashMap<UfeHex, UfeHex>")]
+    pub nonces: HashMap<FieldElement, FieldElement>,
```
xJonathanLEI, line 32: Looks like we're missing a test case where this field is not empty. Am looking for a block with nonce changes and will come back shortly.

starknet-core/src/types/trace.rs
```diff
@@ -19,6 +19,12 @@ pub struct BlockTraces {
 pub struct TransactionTrace {
     /// An object describing the invocation of a specific function.
     pub function_invocation: FunctionInvocation,
+    /// An object describing the invocation of a fee transfer.
+    #[serde(default)]
+    pub fee_transfer_invocation: Option<FunctionInvocation>,
+    /// An object describing the invocation of validation.
+    #[serde(default)]
+    pub validate_invocation: Option<FunctionInvocation>,
```
xJonathanLEI, line 27: Also missing a test case where this field exists. I doubt we can easily find a transaction with this though...

starknet-core/src/types/transaction.rs
xJonathanLEI: We actually had this dicussion before in this thread:

https://github.com/Linear-finance/linear-exchange/pull/27

I think it makes sense for mainnet users to use an older version (a previous commit), especially before we have a fully production-ready mainnet, after which we can start using feature branches for matching testnets.

The thing is attempting to support multiple StarkNet versions inside a single SDK version might not always work, at least not without tradeoffs (in this case, testnet clients are unnecessarily unwrapping `version`).

xJonathanLEI:
Nah don't bother. I think I have access to edit this PR? I have all the suggested changes locally that I will be pushing in a moment :)

starknet-core/src/types/transaction.rs
xJonathanLEI: I think you can actually use different versions of the same crate inside a program, which would solve the issue of fetching mainnet and testnet data inside the same application. IMO this use case is a bit niche, but there is this reasonable workaround.

Things should get much better when the gateway API and JSON-RPC spec stablizes more.

PR #225, opened 2022-10-18
Update starknet 0.10.1 compatibility

fracek:
The testnet gateway was upgraded today. This PR includes the changes to the response to work with the new gateway.

Notice this is a breaking change because of the new deploy account transaction type.


xJonathanLEI (approved):
LGTM. Thanks for working on this!

PR #233, opened 2022-11-11
feat: support testnet 2

xJonathanLEI:
Due to the congestion on the original testnet, a new testnet with chain ID `SN_GOERLI2` has been launched. This PR:

- adds a new const `TESTNET2` the new chain ID; and
- adds a new constructor `starknet_alpha_goerli_2()` to `SequencerGatewayProvider`.

### DIFF
--- a/starknet-core/src/chain_id.rs
+++ b/starknet-core/src/chain_id.rs
@@ -13,3 +13,31 @@ pub const TESTNET: FieldElement = FieldElement::from_mont([
     18446744073709551615,
     398700013197595345,
 ]);
+
+pub const TESTNET2: FieldElement = FieldElement::from_mont([
+    1663542769632127759,
+    18446744073708869172,
+    18446744073709551615,
+    33650220878420990,
+]);
+
+#[cfg(test)]
+mod test {
+    use crate::utils::cairo_short_string_to_felt;
+
+    use super::*;
+
+    #[test]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
+    fn test_chain_ids() {
+        for (text, felt) in [
+            ("SN_MAIN", MAINNET),
+            ("SN_GOERLI", TESTNET),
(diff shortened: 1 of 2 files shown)


milancermak:
Doesn't chain ID stay the same for `alpha-goerli2`?

xJonathanLEI:
@milancermak Wait is it the same? I requested the chain ID from the `pathfinder` synced with `--testnet2`, but I haven't tested calculating transaction hashes. Let me test now.

milancermak:
The CLI in the 0.10.2-pre release [uses the same](https://github.com/starkware-libs/cairo-lang/commit/f951a669e264796d53bf7a3cfd23d5bc1f6e54b8#diff-b329f63899d2303408d748a52452bbe8db173983e573262215afad8d0ecdc8be) (had to dug it up, check the `starknet_cli.py` file if the link doesn't work properly).

xJonathanLEI:
@milancermak You're right, I tried using the "new" chain ID and the signature verification fails on my account. The old one works.

xJonathanLEI:
I guess this means _this_ is wrong then?

https://github.com/eqlabs/pathfinder/blob/7f9a6bb0264943f93a633f61fc4e0bc9237f68a0/crates/pathfinder/src/core.rs#L422

@Mirko-von-Leipzig

xJonathanLEI:
I see this was previously "fixed" in https://github.com/eqlabs/pathfinder/pull/716 but then revereted on purpose in https://github.com/eqlabs/pathfinder/pull/724, citing:

> this is StarkNet bug and not a pathfinder one

@Mirko-von-Leipzig Care to elaborate? Thanks in advance :)

xJonathanLEI:
Thanks for the input @Mirko-von-Leipzig!

I guess I'll just remove the chain ID change from this PR until they actually "fix" it. This is an unfortuante situation though.

milancermak:
Live view of @xJonathanLEI

PR #243, opened 2022-11-19
refactor: move curve operations to new subcrate

xJonathanLEI:
This PR moves Stark curve operations to a new subcrate `starknet-curve` and removes unused curve points.

Before the removal, the wasm output from the `starknet-wasm` example was 328 KB. With this PR the output size becomes 292 KB - 11% reduction.


xJonathanLEI:
Merging anyways since the CI failures are rather caused by #244.

Issue #244, opened 2022-11-19
Support JSONRPC spec 0.2.1


xJonathanLEI:
Support [0.2.1](https://github.com/starkware-libs/starknet-specs/releases/tag/v0.2.1) of the JSON RPC specs now that `pathfinder` supports it.

Issue #232, opened 2022-11-01
Revamp contract deployment


xJonathanLEI:
## Introduction

A lot has changed regarding StarkNet contract deployment since we first implemented `ContractFactory` for deployment contracts:

- Transaction type `Declare` v0 was introduced
- Syscall `deploy` was added to deploy declared classes
- Transaction type `Declare` v1 was introduced, with v0 deprecated
- Transaction type `DeployAccount` was introduced, with `Deploy` deprecated
- The [Universal Deployer](https://community.starknet.io/t/universal-deployer-contract-proposal) contract was proposed and deployed (by @martriay from OpenZeppelin)

So, it's time for a revamp of the contract deployment process in `starknet-rs` to make it more streamlined and robust. Hopefully this document will serve to explain the design decision behind, and to help make things around StarkNet contract deployment clearer as they're kinda messy right now with everything (new & deprecated) floating around.

## How contract deployment works

_Note that at the moment, `starknet-rs` is not interested in supporting deprecated features (since they will be removed after regenesis anyways). Raw data types for sending `Deploy` and `Declare` v0 transaction types will be kept until they're removed from the network itself, but higher-level constructs won't make use of them. Here **we pretend all deprecated features are gone already**._

The contract deployment process consists of 2 steps:

### Declaration

To deploy a contract, you first need to declare its class. The only way to do that is to send a `Declare` v1 transaction _from an account contract_. This is also the only step where you would need the full contract definition. You only need the class hash for the actual deployment.

### Deployment

This is the interesting part: there are actually 3 ways of deploying contracts:

1. Using the `deploy` syscall inside an `Invoke` v1 transaction. However, this is equivalent to `CREATE` in Ethereum, which is mostly none of an SDK's business.
2. Using the [Universal Deployer](https://community.starknet.io/t/universal-deployer-contract-proposal) contract. This is technically the same thing as method 1, but it's "standardized" (well, by OpenZeppelin, so maybe @martriay can shed some light on the level of commitment on interface and behavior) so we can kind of rely on it and code it into `starknet-rs`. This needs to go through an account contract.
3. Using the `DeployAccount` transaction type, where transaction fees are pre-funded to the target contract address. This will likely be the main way of account contract deployment going forward, so we definitely need to support it. This also needs to go through a (future) account contract.

[...]

PR #247, opened 2022-11-22
release: bump starknet-crypto to 0.2.0 (and deps)

xJonathanLEI:
The `starknet-crypto` crate version available on crates.io right now is around 100x slower than the current `master` head. It's time to release a new version so that projects stuck with crates.io (e.g. they also want to publish on crates.io) can take advantage of the performance gains.

- `starknet-ff`: `0.1.0` -> `0.2.0` - breaking changes on an error type
- `starknet-curve`: `0.1.0` - new crate
- `starknet-crypto-codegen`: `0.1.0` - new crate
- `starknet-crypto`: `0.1.0` -> `0.2.0` - there's zero breaking change on the lib itself, but since it re-exports `starknet-ff`, its public API has technically changed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1510,7 +1510,7 @@ dependencies = [
 
 [[package]]
 name = "starknet-crypto"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "criterion",
  "crypto-bigint",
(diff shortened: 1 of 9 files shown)


xJonathanLEI:
Will release new version for other crates after #232 and #244.

PR #251, opened 2022-11-23
feat: support jsonrpc 0.2.1

Matthew-Herman:
These changes add support for the jsonrpc 0.2.1 spec for [issue 244](https://github.com/xJonathanLEI/starknet-rs/issues/244).

### DIFF
--- a/starknet-providers/src/jsonrpc/mod.rs
+++ b/starknet-providers/src/jsonrpc/mod.rs
@@ -111,8 +111,10 @@ struct FeltArray(#[serde_as(as = "Vec<UfeHex>")] pub Vec<FieldElement>);
 struct EventFilterWithPage {
     #[serde(flatten)]
     filter: EventFilter,
-    page_size: u64,
-    page_number: u64,
+    /// a pointer to the last element of the delivered page, use this token in a subsequent query to obtain the next page
+    #[serde(skip_serializing_if = "Option::is_none")]
+    continuation_token: Option<String>,
+    chunk_size: u64,
 }
 
 impl<T> JsonRpcClient<T> {
@@ -224,11 +226,15 @@ where
     /// Get the contract class definition associated with the given hash
     pub async fn get_class(
         &self,
+        block_id: &BlockId,
         class_hash: FieldElement,
     ) -> Result<ContractClass, JsonRpcClientError<T::Error>> {
         self.send_request(
             JsonRpcMethod::GetClass,
-            [serde_json::to_value(Felt(class_hash))?],
+            [
+                serde_json::to_value(block_id)?,
+                serde_json::to_value(Felt(class_hash))?,
+            ],
         )
         .await
     }
@@ -301,18 +307,15 @@ where
     }
 
     /// Estimate the fee for a given StarkNet transaction
-    pub async fn estimate_fee<R>(
+    pub async fn estimate_fee(
         &self,
-        request: R,
+        request: &BroadcastedTransaction,
         block_id: &BlockId,
-    ) -> Result<FeeEstimate, JsonRpcClientError<T::Error>>
-    where
-        R: AsRef<FunctionCall>,
-    {
+    ) -> Result<FeeEstimate, JsonRpcClientError<T::Error>> {
         self.send_request(
             JsonRpcMethod::EstimateFee,
             [
-                serde_json::to_value(request.as_ref())?,
+                serde_json::to_value(request)?,
                 serde_json::to_value(block_id)?,
             ],
         )
@@ -357,15 +360,15 @@ where
     pub async fn get_events(
         &self,
         filter: EventFilter,
-        page_size: u64,
-        page_number: u64,
+        continuation_token: Option<String>,
+        chunk_size: u64,
     ) -> Result<EventsPage, JsonRpcClientError<T::Error>> {
         self.send_request(
             JsonRpcMethod::GetEvents,
             [serde_json::to_value(EventFilterWithPage {
                 filter,
-                page_size,
-                page_number,
+                continuation_token,
+                chunk_size,
             })?],
         )
         .await
@@ -374,12 +377,16 @@ where
     /// Get the latest nonce associated with the given address
     pub async fn get_nonce(
         &self,
+        block_id: &BlockId,
         contract_address: FieldElement,
     ) -> Result<FieldElement, JsonRpcClientError<T::Error>> {
         Ok(self
             .send_request::<_, Felt>(
                 JsonRpcMethod::GetNonce,
-                [serde_json::to_value(Felt(contract_address))?],
+                [
+                    serde_json::to_value(block_id)?,
+                    serde_json::to_value(Felt(contract_address))?,
+                ],
             )
             .await?
             .0)
@@ -388,19 +395,11 @@ where
     /// Submit a new transaction to be added to the chain
     pub async fn add_invoke_transaction(
         &self,
-        function_invocation: &FunctionCall,
-        signature: Vec<FieldElement>,
-        max_fee: FieldElement,
-        version: FieldElement,
+        invoke_transaction: BroadcastedInvokeTransaction,
     ) -> Result<InvokeTransactionResult, JsonRpcClientError<T::Error>> {
         self.send_request(
             JsonRpcMethod::AddInvokeTransaction,
-            [
-                serde_json::to_value(function_invocation)?,
-                serde_json::to_value(FeltArray(signature))?,
-                serde_json::to_value(Felt(max_fee))?,
-                serde_json::to_value(Felt(version))?,
-            ],
+            [serde_json::to_value(invoke_transaction)?],
         )
         .await
     }
@@ -408,15 +407,11 @@ where
     /// Submit a new transaction to be added to the chain
     pub async fn add_declare_transaction(
         &self,
-        contract_class: &ContractClass,
-        version: FieldElement,
+        declare_transaction: &BroadcastedDeclareTransaction,
     ) -> Result<DeclareTransactionResult, JsonRpcClientError<T::Error>> {
         self.send_request(
             JsonRpcMethod::AddDeclareTransaction,
-            [
-                serde_json::to_value(contract_class)?,
-                serde_json::to_value(Felt(version))?,
-            ],
+            [serde_json::to_value(declare_transaction)?],
         )
         .await
     }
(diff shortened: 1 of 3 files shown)


starknet-providers/src/jsonrpc/mod.rs
```diff
@@ -111,8 +111,10 @@ struct FeltArray(#[serde_as(as = "Vec<UfeHex>")] pub Vec<FieldElement>);
 struct EventFilterWithPage {
     #[serde(flatten)]
     filter: EventFilter,
-    page_size: u64,
-    page_number: u64,
+    /// a pointer to the last element of the delivered page, use this token in a subsequent query to obtain the next page
```
xJonathanLEI, line 114: No comment/docs lines should exceed 100 chars for readability. Hard-wrap it into the next line. Please also fix other occurrences.

xJonathanLEI (changes requested):

starknet-providers/src/jsonrpc/mod.rs
```diff
@@ -224,11 +226,15 @@ where
     /// Get the contract class definition associated with the given hash
```
xJonathanLEI, line 226: The description for this method has changed to `Get the contract class definition in the given block associated with the given hash` in the specs. Please reflect it here in the docs. Please also check for other desc changes.

starknet-providers/src/jsonrpc/mod.rs
```diff
@@ -301,18 +307,15 @@ where
...
         &self,
-        request: R,
+        request: &BroadcastedTransaction,
         block_id: &BlockId,
-    ) -> Result<FeeEstimate, JsonRpcClientError<T::Error>>
-    where
-        R: AsRef<FunctionCall>,
-    {
+    ) -> Result<FeeEstimate, JsonRpcClientError<T::Error>> {
         self.send_request(
```
xJonathanLEI, line 315: The use of generic here is by design. In some cases it's impossible to pass references since the value gets dropped immediately. Using `AsRef` generic here allows the function to be used in those cases.

Extra caution is needed if you're changing the design.

Matthew-Herman: Reverted to using the generic for estimate_fee

starknet-providers/src/jsonrpc/models/mod.rs
```diff
@@ -591,6 +594,228 @@ pub struct BlockHashAndNumber {
...
+    Declare,
+    Deploy,
+    DeployAccount,
+    Invoke,
+    L1Handler,
+}
+
+/// (`CONTRACT_ABI`)
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ContractABI {
```
xJonathanLEI, line 731: Rename all `ABI` to `Abi`

starknet-providers/src/jsonrpc/mod.rs
```diff
@@ -388,53 +400,35 @@ where
...
             JsonRpcMethod::AddDeployTransaction,
-            [
-                serde_json::to_value(Felt(contract_address_salt))?,
-                serde_json::to_value(FeltArray(constructor_calldata))?,
-                serde_json::to_value(contract_definition)?,
-            ],
+            [serde_json::to_value(deploy_transaction)?],
         )
         .await
     }
```
xJonathanLEI, line 434: Missing the new `starknet_addDeployAccountTransaction` method?

starknet-providers/src/jsonrpc/models/mod.rs
```diff
@@ -286,14 +284,15 @@ pub struct StorageDiffItem {
     pub value: FieldElement,
 }
 
-/// Transaction (`TXN`)
+/// Transaction (`TXN`) The transaction schema, as it appears inside a block
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(tag = "type", rename_all = "SCREAMING_SNAKE_CASE")]
+#[serde(untagged)]
```
xJonathanLEI, line 289: The change to `untagged` here seems unjustified. It shouldn't be necessary for making things work, since the raw data transferred over the wire _does_ come tagged.

starknet-providers/src/jsonrpc/models/mod.rs
```diff
@@ -396,47 +388,64 @@ pub struct TransactionReceiptMeta {
...
-pub struct DeclareOrDeployTransactionReceipt {
+pub struct L1HandlerTransactionReceipt {
     #[serde(flatten)]
     pub meta: TransactionReceiptMeta,
 }
 
-/// The `TXN_RECEIPT` type in the specification, except without the `PENDING_TXN_RECEIPT` variant.
+/// (`TXN_RECEIPT`)
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(untagged)]
```
xJonathanLEI, line 441: This `untagged` used to be necessary, but it looks like the new spec version does tag the receipts, making the use of `untagged` here unnecessary. Change it to tagged.

xJonathanLEI: `untagged`s on all receipt types should be reviewed as they should no longer be necessary.

starknet-providers/src/jsonrpc/models/mod.rs
```diff
@@ -591,6 +594,228 @@ pub struct BlockHashAndNumber {
...
+
+/// (`CONTRACT_ABI`)
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ContractABI {
+    pub entries: Vec<ContractABIEntry>,
+}
+
+/// (`CONTRACT_ABI_ENTRY`)
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
```
xJonathanLEI, line 737: This should not be untagged. It's tagged in the spec.

xJonathanLEI (changes requested):
Apart from the comments already made, looks like `untagged` is misused across the PR. Please review the use of `untagged` and change accordingly.

Matthew-Herman:
Tried to remove use of untagged where possible. Not sure if there is a way to avoid using `.clone()` in the add transaction methods? I added it because initialising the enum there to add the tag takes ownership of the method param.

fracek:
FYI, this patch is needed to use the json rpc provider with infura. Any way I can help to get this merged?

xJonathanLEI:
@fracek ima get it done tmr

Matthew-Herman:
Removed the tests for add_deployTransaction because StarkNet v0.10.3 removes the deploy transaction and was deployed to goerli yesterday. Also removed tests for adding version 0 transactions because they are deprecated on the docs.

starknet-providers/src/jsonrpc/models/mod.rs
```diff
@@ -18,21 +18,15 @@ pub enum MaybePendingBlockWithTxHashes {
     PendingBlock(PendingBlockWithTxHashes),
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Deserialize)]
```
xJonathanLEI, line 21: There are many occurrences where the `Serialize` derive gets removed by this PR. Please don't do this. Technically we don't need serialization for the JSON-RPC client to work. However, as an SDK we need to consider use cases of dependent libs/apps, who are very likely to need serialization to function properly.

xJonathanLEI: Serialize trait derives added back.

starknet-providers/src/jsonrpc/mod.rs
```diff
@@ -388,53 +393,56 @@ where
...
-                serde_json::to_value(FeltArray(signature))?,
-                serde_json::to_value(Felt(max_fee))?,
-                serde_json::to_value(Felt(version))?,
-            ],
+            [serde_json::to_value(BroadcastedTransaction::Invoke(
+                invoke_transaction.clone(),
+            ))?],
         )
         .await
     }
```
xJonathanLEI, line 405: Asking the user to pass in reference only to `clone()` inside is bad practice. The user might have an object he's willing to give up ownership for, but this function here is forcing an extra clone, which could be expensive.

The responsibility for cloning is on the user, _if_ we have to take ownership.

starknet-providers/src/jsonrpc/models/mod.rs
```diff
@@ -591,6 +681,196 @@ pub struct BlockHashAndNumber {
...
+    pub max_fee: FieldElement,
+    /// Version of the transaction scheme
+    pub version: TransactionVersion,
+    #[serde_as(as = "Vec<UfeHex>")]
+    pub signature: Vec<FieldElement>,
+    #[serde_as(as = "UfeHex")]
+    pub nonce: FieldElement,
+    /// The salt for the address of the deployed contract
+    #[serde_as(as = "UfeHex")]
+    pub contract_address_salt: FieldElement,
```
xJonathanLEI, line 768: I don't like how this allows people to construct invalid transaction requests: you can specify `version` to be `V0` without encountering a compile-time error. However there's actually no v0 for `DeployAccount` transaction type.

starknet-providers/src/jsonrpc/mod.rs
xJonathanLEI: Actually, there's no need to take ownership here at all. You shouldn't need to wrap the object in a `BroadcastedTransaction` enum in the first place. It's just for serialization after all.

Matthew-Herman: Moved tagging with type from the enum to structs as it is only used for serialization

starknet-providers/src/jsonrpc/models/mod.rs
Matthew-Herman: Understand we should do this. I'm a bit confused about which transaction versions the types should allow to be constructed here. Not sure the [docs.starknet.io](https://docs.starknet.io/documentation/architecture_and_concepts/Blocks/transactions/#declare_transaction) and jsonrpc spec are giving me a full picture. Should this just target v0.10.3 and should constructing transaction version 0 for declare/invoke/deploy transactions be allowed? And also now that adding deploy transactions are deprecated from v0.10.3, should the add_deploy_transaction method be removed?

xJonathanLEI: I general we don't support deprecated features since they will be removed from the protocol completely (without needing to support old block data) eventually with the regenesis. However, I want to make an exception here for JSON RPC as we're specifically targeting a certain specification. Making our code closer to the specs helps with maintenance.

xJonathanLEI:
Superseded by #255.

PR #255, opened 2022-12-11
feat(jsonrpc): update to support spec v0.2.1

### LINKED ISSUE #244: Support JSONRPC spec 0.2.1

xJonathanLEI:
Resolves #244.

Supersedes #251.

While the work in #251 by @Matthew-Herman was quite complete, I decided to build on it with a new approach and implement codegen from the specs instead:

- it helps reduce future maintenance effort;
- it helps make sure we're not accidentally deviating from the specs.

The codegen tool is available [here](https://github.com/xJonathanLEI/starknet-jsonrpc-codegen).

Technically we're not directly deriving everything from the specs though, for some reasons:

- the specs itself isn't completely correct: https://github.com/starkware-libs/starknet-specs/pull/64;
- some semantics are not captured by the specs (e.g. the `type` field value is directly linked to the variant, but the specs doesn't declare the relationship);
- some types are too complicated to be automatically generated for them to be idiomatic.

Alternative approaches have been considered:

- proc macros: not exactly a good fit for value objects like this as it makes exploring the source rather difficult;
- `build.rs`: better than proc macros but users still need to dig into the `target` folder for the generated source.

This PR also makes the contract deployment test ignored by default since it's no longer possible to send deploy transactions to live environments.

### DIFF
--- a/starknet-contract/tests/contract_deployment.rs
+++ b/starknet-contract/tests/contract_deployment.rs
@@ -2,6 +2,7 @@ use starknet_contract::ContractFactory;
 use starknet_core::types::{ContractArtifact, FieldElement};
 
 #[tokio::test]
+#[ignore = "Legacy deploy transaction removed from live environments"]
 async fn can_deploy_contract_to_alpha_goerli() {
     let artifact = serde_json::from_str::<ContractArtifact>(include_str!(
         "../test-data/artifacts/oz_account.txt"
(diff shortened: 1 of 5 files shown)


xJonathanLEI:
cc @fracek

xJonathanLEI:
I actually still don't quite like the part where `version` fields are just plain numbers. Will work on it on a later PR but just wanna push something that works first.

PR #264, opened 2022-12-21
refactor: use from_str when deserializing FieldElement

kariy:
This PR refactors `Deserialize` trait of `FieldElement` using `from_str` so that can deserialize both hex and decimal string.


xJonathanLEI:
Thanks for the PR! There seems to be a clippy error. Can you fix it? Thanks.

xJonathanLEI (approved):
LGTM

PR #265, opened 2022-12-31
feat: change default localhost port to 5050 to meet current conventio…

ponderingdemocritus:
Updates port value to `5050` to conform with the rest of the ecosystem standards (nile, devnet etc)

Only a minor change - but will stop users from having to have custom ports for the other services.


xJonathanLEI (approved):
LGTM. Thanks for the PR. I don't use these tools myself so was unaware of the default port used.

PR #266, opened 2023-01-06
feat: introduce starknet-id support

Th0rgal:
Hey there,

This pull request brings .stark names support natively in a trustless way. Contract calls are done through the provided gateway and the encoding/decoding has been rewritten in Rust from https://docs.starknet.id/for-devs/encoding-algorithm.

I added a macro for easy use on mainnet (it relies on ``SequencerGatewayProvider::starknet_alpha_mainnet()`` and ``MAINNET_CONTRACT``):
```rs
let addr = domain_to_address!("th0rgal.stark").await.unwrap();
let domain = address_to_domain!(FieldElement::from_hex_be(
        "0x048F24D0D0618fa31813DB91a45d8be6c50749e5E19ec699092CE29aBe809294"
    )
    .unwrap())
    .await;
```

The encode and decode functions are also exposed publicly. They could be useful for more complex interactions with the contract or simply to be able to store strings efficiently on a contract.

I added examples and tests on my domain, I don't know if I should leave them like that because they could break if the starknet endpoint changes or if I change my main domain (which I shouldn't do anytime soon).

I hope this will help, let me know if I can improve anything,
Thomas


xJonathanLEI:
CI red :/

xJonathanLEI:
Thanks for the PR! On top of the red CI, I'm actually not very happy about how the design. I think this functionality should not live in a separate subcrate, but instead `starknet-core` itself, if we want to eventually support using it for everything.

To be more specific, this PR is basically two parts:

1. Encoding/decoding: this should probably live in `starknet-core` as util functions.
2. Resolving: in this PR it's implemented as util functions/macros, but I think a better apporach would be to do it in a `impl<P: Provider> for P` block in `starknet-providers`. This should be more idiomatic IMO.

I think the ENS support in `ethers-rs` is very nice. Maybe it's worth taking a look.

Th0rgal:
> Thanks for the PR! On top of the red CI, I'm actually not very happy about how the design. I think this functionality should not live in a separate subcrate, but instead `starknet-core` itself, if we want to eventually support using it for everything.
> 
> To be more specific, this PR is basically two parts:
> 
> 1. Encoding/decoding: this should probably live in `starknet-core` as util functions.
> 2. Resolving: in this PR it's implemented as util functions/macros, but I think a better apporach would be to do it in a `impl<P: Provider> for P` block in `starknet-providers`. This should be more idiomatic IMO.
> 
> I think the ENS support in `ethers-rs` is very nice. Maybe it's worth taking a look.

You are welcome. I think the red CI is just a formatting issue, it should be fixed now. Encoding/Decoding is quite specific to starknet-id (it is not UTF-8 or anything standard) so I'm not sure you want to include it in starknet-core. The suggestion for starknet-providers looks great, I'll check out ethers-rs. :)

Th0rgal:
up @xJonathanLEI

xJonathanLEI:
After a second thought, I believe the SDK should stay neutral and not have deep integration with any protocol. Starknet ID name resolution can be easily implemented as a third party library that depends on `starknet-rs`. Closing this PR now.

PR #267, opened 2023-01-09
feat: optional feature no_unknown_fields

xJonathanLEI:
This PR adds a new optional feature `no_unknown_fields`. This feature is mainly for our tests and not exactly intended for production use, as it could break your deployment when the server upgrades.

This change is made because there's no way to pass `#[cfg(test)]` to integration tests, which we previously used to conditonally set `deny_unknown_fields`. As a result, there was no way to test our JSON-PRC client with that turned on in integration tests.


xJonathanLEI:
`FunctionAbiEntry` is intentionally left out until this issue resolves:

https://github.com/eqlabs/pathfinder/issues/814

PR #269, opened 2023-01-14
feat: add deploy contract with udc example

omarespejel:
This PR adds a new example; it shows how to deploy a contract using the UDC.

This change is made because the current [deployment example](https://github.com/xJonathanLEI/starknet-rs/blob/master/examples/deploy_contract.rs) doesn't work because the "deploy" transaction is deprecated .

### DIFF
--- a/examples/deploy_contract_udc.rs
+++ b/examples/deploy_contract_udc.rs
@@ -1,0 +1,57 @@
+use starknet::{
+    accounts::{Account, Call, SingleOwnerAccount},
+    core::{
+        chain_id,
+        types::{ContractArtifact, FieldElement},
+        utils::get_selector_from_name,
+    },
+    providers::SequencerGatewayProvider,
+    signers::{LocalWallet, SigningKey},
+};
+
+#[tokio::main]
+async fn main() {
+    // Defining an account contract in the Testnet
+    let provider = SequencerGatewayProvider::starknet_alpha_goerli();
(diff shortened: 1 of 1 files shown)


xJonathanLEI:
Hi! Sorry for merging my own newer PR (#271) first, which contains breaking changes and this PR doesn't work anymore. Could you please rebase `master` and:

- instead of adding the example, replace the current useless deployment example; and
- showcase the use of the `felt!()` macro for constructing `FieldElement`s instead of doing the unwraps?

Thanks!

xJonathanLEI:
Of if you don't mind, I can just amend your commit directly :)

omarespejel:
Hey @xJonathanLEI! Sure, I can proceed with both points you mention :)

Will work on that this week. Congratulations on the revamping of the `Account` trait!

omarespejel:
I think that, after merging PR #279, the two points mentioned above would not longer be necessary. WDYT @xJonathanLEI?

Unless we want to explicitly show how deployment can be done directly by invoking the UDC.

xJonathanLEI:
As part of implementing #279 I had to also change the example code so that it compiles, effectively superseding this PR. I'm sorry but I guess I need to close this one now @omarespejel.

Thanks for the PR regardless!

omarespejel:
No problem at all! I will go on and work on #5.

PR #301, opened 2023-02-05
chore: bump `ark-ff` to 0.4.2

xJonathanLEI:
This PR bumps `ark-ff` to 0.4.0 and resolves all the breaking changes.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -101,23 +101,23 @@ jobs:
 
       - name: Run starknet-ff tests
         run: |
-          (cd ./starknet-ff && wasm-pack test --node)
+          (cd ./starknet-ff && wasm-pack test --release --node)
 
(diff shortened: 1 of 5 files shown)


xJonathanLEI:
Bumping `ark-ff` to 0.4.0 seems to cause the wasm module generated under debug mode to use more than 50,000 locals in a single function, causing the wasm tests to fail. If we merge this, it's gonna have the unwanted side effect of forcing downstream applications to also run their wasm tests in release mode.

I'm putting this PR as draft until a better solution exists.

xJonathanLEI:
Just tried benchmarking a bit and it looks like the upgrade actually causes a performance drop of around 5-10% across the board... Not sure if we should still upgrade.

xJonathanLEI:
Whatever... let's just force every downstream application to test their wasm builds in release mode.

PR #303, opened 2023-02-13
Add no_std compatibility

abdelhamidbakhta:
# Description

This PR enables `no_std` compatibility for `starknet-rs`.
It uses [no-std-compat](https://crates.io/crates/no-std-compat) crate to achieve this goal.

# Usage

If you want to use a crate from `starknet-rs` with `no-std` support you can do like this:

```toml
starknet-crypto = { git = "...", rev = "...", default-features = false }

std = [
  "starknet-crypto/std",
]
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1668,6 +1668,8 @@ name = "starknet-ff"
 version = "0.2.0"
 dependencies = [
  "ark-ff",
+ "ark-serialize",
+ "ark-std",
  "bigdecimal",
  "crypto-bigint",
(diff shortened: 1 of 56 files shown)


abdelhamidbakhta:
Actually it seems we won't be able to use `no-std-compat`.

We will likely have to handle the compatibility manually, like what Substrate std is doing: 
https://github.com/paritytech/substrate/blob/polkadot-v0.9.37/primitives/std/src/lib.rs

PR #304, opened 2023-02-16
Add `no_std` compatibility

abdelhamidbakhta:
# Description

This PR enables `no_std` compatibility for `starknet-rs`.

# Usage

If you want to use a crate from `starknet-rs` with `no-std` support you can do like this:

```toml
starknet-crypto = { git = "...", rev = "...", default-features = false }

std = [
  "starknet-crypto/std",
]
```

The compatibility required to use:
- `thiserror-no-std` instead of `thiserror`
- fork `bigdecimal` to add `no_std` support and publish a `bigdecimal-no-std` crate


xJonathanLEI:
Does this one supersede #303?

abdelhamidbakhta:
> Does this one supersede #303?

Yes I closed the other one

xJonathanLEI:
The wasm CI is red. I'm not very familiar with `no_std` here - do we need to change how wasm tests are executed?

abdelhamidbakhta:
> The wasm CI is red. I'm not very familiar with `no_std` here - do we need to change how wasm tests are executed?

I will investigate why wasm tests are not passing, meanwhile I switch the PR as draft until I find the cause

PR #306, opened 2023-02-22
feat: introduce no_std support for ff, curve, and crypto

tdelabro:


starknet-crypto/src/lib.rs
```diff
@@ -1,5 +1,24 @@
...
+#[cfg(feature = "std")]
+include!("./with_std.rs");
+
+#[cfg(not(feature = "std"))]
+include!("./without_std.rs");
+
+#[cfg(all(test, not(feature = "std"), feature = "alloc"))]
+include!("./with_alloc.rs");
+
+pub mod stdlib {
```
xJonathanLEI, line 13: Do we need this to be `pub`? I think we shouldn't re-export it?

starknet-crypto/src/with_std.rs
```diff
@@ -0,0 +1,6 @@
+pub mod with_std {
+    pub use std::{fmt, ops, str};
```
xJonathanLEI, line 2: The global `core` crate is available even when `no_std` is off (aka. normal mode). So we should not need this conditional import? Should be able to just change all `std::fmt` etc. to `core::fmt` unconditionally for example?

xJonathanLEI:
Found this clippy lint that might be useful:

https://github.com/rust-lang/rust-clippy/issues/7724

starknet-crypto/src/with_std.rs
xJonathanLEI: I think the best approach here is to apply the clippy lint(s) project-wide in `master` first, and rebase this PR? That way we don't have to manually inspect every conditional import to see if they're not needed.

xJonathanLEI: Played around the idea a bit and got bitten by `thiserror` lol - enabling the lints results in every `thiserror::Error` derive getting error. So I would have to deal with `thiserror` first, which is exactly what this PR already does.

So I'll simply add the lints on top of this PR to `starknet-core` and its dependencies to make sure we're not doing unnecessary conditional imports. Then I will apply the lints to other crates in a separate PR.

(btw I'm kinda annoyed that there's no way to enable/disable lints globally in a workspace/project, not until this merges: https://github.com/rust-lang/rfcs/pull/3389)

xJonathanLEI:
Note that we should also add CI steps for building without the `std` feature to make sure we're not accidentally breaking `no_std` support.

xJonathanLEI:
Hmmm I just tried running this:

```console
$ cargo build --package starknet-crypto --no-default-features
```

to make sure it _really_ works for `no_std`, but I'm getting errors. Any ideas?

xJonathanLEI:
Is it possible that we make the crate work with everything turned off (not even `alloc`), but with reduced functionality? Anything that requires `alloc` gets gated by `any(feature = "std", feature = "alloc")`. Then at least the crate would work with a bare `default-features = false`.

tdelabro:
> Is it possible that we make the crate work with everything turned off (not even alloc), but with reduced functionality?

Yeah totally. `alloc` is only used by starknet-ff here:
```rust
impl Serialize for FieldElement {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: serde::Serializer,
    {
        serializer.serialize_str(&ToString::to_string(&self))
    }
}

impl<'de> Deserialize<'de> for FieldElement {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let value = String::deserialize(deserializer)?;
        Self::from_str(&value).map_err(serde::de::Error::custom)
    }
}
```

So what we can do is:
```toml
[dependencies]
serde = { version = "1.0.152", default-features = false, optional = true }

[features]
serde = ["alloc", "dep:serde"]
alloc = [""serde?/alloc"]
```

starknet-crypto/src/with_std.rs
tdelabro: > Should be able to just change all std::fmt etc. to core::fmt unconditionally for example?

Yes we can. But then we have to do conditional import for symbols in `alloc`:
```rust
#[cfg(feature = "std")]
use std::rc::Rc;
#[cfg(all(not(feature = "std")), feature = "alloc")]
use alloc::rc::Rc;
```

I tried both and I really find it more comfortable to have all those stuffs managed once and for all in one place, by the library maintainer, and then anybody contributing just has to import what it needs from one single place.

It may seem a bit artificial for small repositories that don't do extensive usage of the `std`/`core`/`alloc` symbols, but in bigger repositories, it is such a quality-of-life improvement.

Nevertheless, I'm open to removing it and keeping it manual if it really bother you

tdelabro: About `thiserror`. It does not support `no_std`. I used `this-error-no-std` instead, which is kind of a wrapper crate I believe. With a lot of complex macro in it, so it is possible that it does strange stuffs to the linter
`snafu` is an alternative crate that supports `no_std` natively, but it is not iso-feature with `thiserror`. It's another paradigm.

starknet-crypto/src/lib.rs
tdelabro: yep you are right

tdelabro:
@xJonathanLEI I pushed a new version. Where the tree crates are full `#![no_std]` and don't relly at all on `std`. 
They occasionally need `alloc` to perform some specific actions that are behind feature flags, such as `serde` or `display-field-element`.

I also replaced `thiserror` by `snafu` in order to have non-buggy `no_std`.

I think you will like this version better

starknet-crypto/Cargo.toml
```diff
@@ -10,22 +10,29 @@ homepage = "https://starknet.rs/"
...
+starknet-ff = { version = "0.2.0", path = "../starknet-ff", default-features = false }
+crypto-bigint = { version = "0.4.9", default-features = false }
+hmac = { version = "0.12.1", default-features = false }
+num-bigint = { version = "0.4.3", default-features = false }
+num-integer = { version = "0.1.44", default-features = false }
+num-traits = { version = "0.2.14", default-features = false }
+rfc6979 = { version = "0.3.1", default-features = false }
+sha2 = { version = "0.10.6", default-features = false }
+zeroize = { version = "1.5.0", default-features = false }
+hashbrown = { version = "0.13.2", default-features = false }
```
xJonathanLEI, line 27: Is there a specific reason `hashbrown` gets added here?

starknet-ff/Cargo.toml
```diff
@@ -10,26 +10,29 @@ homepage = "https://starknet.rs/"
...
-crypto-bigint = "0.4.9"
-hex = "0.4.3"
-num-bigint = { version = "0.4.3", optional = true }
-serde = "1.0.152"
-thiserror = "1.0.30"
+ark-ff = { version = "0.3.0", default-features = false }
+crypto-bigint = { version = "0.4.9", default-features = false }
+hex = { version = "0.4.3", default-features = false }
+serde = { version = "1.0.152", optional = true, default-features = false }
+bigdecimal-no-std = { version = "0.4.0", optional = true, default-features = false }
```
xJonathanLEI, line 23: As mentioned in the Telegram chat, it looks like `bigdecimal-no-std` isn't _really_ `no_std` compatible as its name suggests - I still have to enable `std` for it to compile for `thumbv6m-none-eabi`.

Maybe we simply revert back to using the original `bigdecimal` crate and make it depend on `std` instead?

starknet-ff/Cargo.toml
```diff
@@ -10,26 +10,29 @@ homepage = "https://starknet.rs/"
...
+num-bigint = { version = "0.4.3", optional = true, default-features = false }
+
+
+[features]
+default = ["std"]
```
xJonathanLEI, line 28: I think it makes more sense for `serde` to also be enabled by default.

xJonathanLEI (commented):
Thanks for updating this! Looks good to me apart from a few minor issues

starknet-crypto/Cargo.toml
tdelabro: nope. It's an error

starknet-ff/Cargo.toml
tdelabro: ok

starknet-ff/Cargo.toml
tdelabro: > Maybe we simply revert back to using the original bigdecimal crate and make it depend on std instead?

Done. Good enough for now.
I will try to create a `no_std` branch for `bigdecimal` later, and maybe merge it in the original repo. But it is not really active...

starknet-ff/src/lib.rs
```diff
@@ -1,14 +1,19 @@
+#![cfg_attr(not(feature = "std"), no_std)]
 #![allow(clippy::comparison_chain)]
 
-use crate::fr::FrParameters;
+#[cfg(all(not(feature = "std"), any(test, feature = "alloc")))]
+#[cfg_attr(test, macro_use)]
+extern crate alloc;
+
+pub use core::{
```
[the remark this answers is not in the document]
tdelabro: totally

xJonathanLEI (approved):
LGTM. Thanks a lot for working on this!

Issue #286, opened 2023-01-25
Parsing a `FunctionCall` call return variable of type`Vec<FieldElement>` using a contract's ABI


Eikix:
Hey:), 

I'm currently using `starknet-rs` to call the [Kakarot contract](https://github.com/sayajin-labs/kakarot). Specifically, its entrypoint named `execute_at_address` (although it is an external function, I call it here as a view).

Here is the signature of the function:
```cairo
func execute_at_address{
    syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr, bitwise_ptr: BitwiseBuiltin*
}(address: felt, value: felt, gas_limit: felt, calldata_len: felt, calldata: felt*) -> (
    stack_accesses_len: felt,
    stack_accesses: felt*,
    stack_len: felt,
    memory_accesses_len: felt,
    memory_accesses: felt*,
    memory_bytes_len: felt,
    evm_contract_address: felt,
    starknet_contract_address: felt,
    return_data_len: felt,
    return_data: felt*,
)
```

I manage to call this function fine, and the return is:
```
[...]

PR #318, opened 2023-03-02
[Feature]: decode a slice of `FieldElement` into a list of `Token`

haroune-mohammedi:
This PR addresses issue #286

We implemented a `decode` fucntion
```rust
fn decode(types: &[ParamType], data: &[FieldElement]) -> Result<Vec<Token>, DecoderError>
```

Where:
```rust
pub enum ParamType {
    FieldElement,
    Array,
    Tuple(usize),
}

#[derive(PartialEq, Eq, Debug)]
pub enum Token {
    FieldElement(FieldElement),
    Array(Vec<FieldElement>),
    Tuple(Vec<FieldElement>),
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1172,18 +1172,18 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.51"
+version = "1.0.52"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5d727cae5b39d21da60fa540906919ad737832fe0b1c165da3a34d6548c849d6"
+checksum = "1d0e1ae9e836cc3beddd63db0df682593d7e2d3d891ae8c9083d2113e1744224"
 dependencies = [
  "unicode-ident",
 ]
 
 [[package]]
 name = "quote"
-version = "1.0.18"
+version = "1.0.26"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1feb54ed693b93a84e14094943b84b7c4eae204c512b7ccb95ab0c66d278ad1"
+checksum = "4424af4bf778aae2051a77b60283332f386554255d722233d09fbfc7e30da2fc"
 dependencies = [
  "proc-macro2",
 ]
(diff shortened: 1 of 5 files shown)


fracek:
I like where you're going with this PR. My suggestion is to add a new trait for encoding and decoding Starknet types.

```rs
pub trait AbiEncode {
  /// Encode type to a token.
  // TODO: should it consume the type or should take a reference?
  fn encode(self) -> Token;
}

pub trait AbiDecode {
  type Error;
  
  fn decode(tokens: impl AsRef<[Token]>) -> Result<Self, Self::Error>;
}
```

Then it's possible to implement the encoder and decoder for all rust builtin types like numbers (`u64`, `u32`), and structs (vectors, tuples, and arrays).

haroune-mohammedi:
@xJonathanLEI can you please take a look here and tell us what do you think ?

I added a macro that auto generates a `Decode` implementation for structs from a slice of `Token`, check https://github.com/xJonathanLEI/starknet-rs/blob/59f42eaddab7651e94474cd70f8872b27eadebea/examples/decoder_example.rs

haroune-mohammedi:
> I like where you're going with this PR. My suggestion is to add a new trait for encoding and decoding Starknet types.
> 
> ```rust
> pub trait AbiEncode {
>   /// Encode type to a token.
>   // TODO: should it consume the type or should take a reference?
>   fn encode(self) -> Token;
> }
> 
> pub trait AbiDecode {
>   type Error;
>   
>   fn decode(tokens: impl AsRef<[Token]>) -> Result<Self, Self::Error>;
> }
> ```
> 
> Then it's possible to implement the encoder and decoder for all rust builtin types like numbers (`u64`, `u32`), and structs (vectors, tuples, and arrays).

I wen with `TryFrom` for converting `Token` to rust types because that's what is used for `FieldElement`

Now, after I introduced `Decode` to convert a `Vec<Token>` to a struct, naming the trait that convert `Token` to a "scalar" type `AbiDecode` will be confusing, maybe using `TryFrom` is the best option after all, what do you guys think ?

glihm:
@haroune-mohammedi @fracek it could be awesome to have your feedbacks on #475 related to this! As this one may be closed due to the overlap. :+1:

PR #322, opened 2023-03-10
feat: revert breaking change on `Signature`

xJonathanLEI:
Undoes the addition of the `v` field to `Signature` introduced in #315 by @greged93.

We don't want the `v` field in `Signature` because the same struct is also used by `starknet-core` for signature verification, which does not make use of `v`. The addition of the field forces users to supply a `v` value for that.

A new struct `ExtendedSignature` is added instead, which can be converted into `Signature`.

### DIFF
--- a/starknet-crypto/src/ecdsa.rs
+++ b/starknet-crypto/src/ecdsa.rs
@@ -22,12 +22,42 @@ pub struct Signature {
     pub r: FieldElement,
     /// The `s` value of a signature
     pub s: FieldElement,
+}
+
+/// Stark ECDSA signature with `v`
+#[derive(Debug)]
+pub struct ExtendedSignature {
+    /// The `r` value of a signature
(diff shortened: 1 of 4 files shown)


starknet-crypto/src/ecdsa.rs
```diff
@@ -38,6 +56,18 @@ impl core::fmt::Display for Signature {
     }
 }
 
+#[cfg(feature = "signature-display")]
+impl core::fmt::Display for ExtendedSignature {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(
+            f,
+            "{}{}",
```
milancermak, line 64: Shouldn't it also display `v`?

xJonathanLEI: Yeah you're right. I overlooked this lol. Thanks

PR #328, opened 2023-03-14
docs: add C++ example with cxx

### LINKED ISSUE #325: Add a new example for showcasing using the library from C++
It's possible to use the library from C++ with [cbindgen](https://github.com/eqrion/cbindgen) or [cxx](https://github.com/dtolnay/cxx). There's probably interest in using it that way, especially for the `starknet-crypto` crate.

We should add an example `starknet-cxx` (something similar to the existing `starknet-wasm` example) that showcases exposing C++ APIs using the [cxx](https://github.com/dtolnay/cxx) bridge.

xJonathanLEI:
Resolves #325.

This PR adds a new example for using `starknet-crypto` from C++.

Given the state of the official C++ crypto library (see https://github.com/starkware-libs/crypto-cpp/issues/2), it probably makes sense to use `starknet-rs` from C++. Adding this example considering that the C++ audience might not be familiar with Rust.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -91,7 +91,7 @@ jobs:
 
       - name: Build project
         run: |
-          cargo build --all --target wasm32-unknown-unknown
+          cargo build --all --target wasm32-unknown-unknown --exclude starknet-cxx
 
       - name: Install wasm-pack
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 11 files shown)


xJonathanLEI:
Looks like it doesn't work for Windows when using the gnu toolchain. Ignored it on CI since it's just an example anyways.

Not sure why it doesn't work though. Not a Windows expert.

Issue #329, opened 2023-03-18
Signature mismatch between Rust and Python versions


rbdm-qnt:
Cont. of this issue: https://github.com/starkware-libs/crypto-cpp/issues/2

I'm trying to get the ECDSA signature generated in Rust to match the one generated by this Python implementation:
https://github.com/starkware-libs/starkex-resources-wip/blob/master/crypto/starkware/crypto/signature/signature.py 

I'm choosing this reference implementation because I've been using it in my application before and it gets verified correctly by the service I'm interfacing with. 

Test:

`
// 2 randomly generated hex strings
msg = "0x1f49a1b3be8f253cdce6b3f996b342c6e8d6ad4c4b4f6f4ab6d8d6b48dd2c7"
key = "0x8ad6c4f4b7b6e4ab6d8b6d48e2c7a6f9b2c8d6e4c6b4f6d4ab6d8d6b48dd2c"

// Python version:
// I'm calling it with no seed, so the extra_entropy gets initialized to b'' as you can see in line 111. Not sure what's the Rust equivalent for this, I'm trying to use an empty string
r, s = sign(int(msg, 16), int(key, 16))

// Rust version, wrapped in C++ using the provided CXX example code:
// I made it output R and S concatenated and separated by a semicolon using format!("{};{}", r_hex, s_hex) as the ecdsa_sign return
auto k = generate_k(msg, key, "");
const char* k_c_str = k.c_str();
std::string k_str(k_c_str);
k_str = "0x" + k_str;
auto signature = ecdsa_sign(msg, key, k_str);
`
Python output:
r: 0x698c5654560a84b16d210692ec4207c09a8ecc21df043284a6382cab8c2079d
s: 0x6386bdd3a021350c8218b68118e275885f8138bb44d9e9f400c69fc510c928e

Rust:
r: 0x698c5654560a84b16d210692ec4207c09a8ecc21df043284a6382cab8c2079d
s: 0x7498bb37722fefb1d3c99b5c70c300a461eeb0d8b2faf943b9198b2ec054f6c

The k values generated by Python's generate_k_rfc6979 and by Rust's generate_k match correctly, and it's this:
0x46e17a0e758084beb8fabd15f070b6cdb77f6d6025501a6d0d5f671f75477f2

I noticed the r values between the 2 implementations always match, and the s value never does.
[...]

PR #330, opened 2023-03-18
docs: use starknet-core for the starknet-cxx example instead

xJonathanLEI:
See #329.

It seems extremely easy for people to shoot themselves in the foot for using `starknet-crypto` directly, especially for developers not familiar with Rust - it's harder for them to inspect the code to see what's _really_ going on.

This PR changes to wrap `starknet-core` instead, and only expose the high-level `ecdsa_sign` function, which takes care of the `k` generation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1657,7 +1657,7 @@ version = "0.1.0"
 dependencies = [
  "cxx",
  "cxx-build",
- "starknet-crypto",
+ "starknet-core",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


xJonathanLEI:
Performance seems to be affected a little bit, with the output binary size also quite a bit larger. This shouldn't be the case as the overhead of wrapping `starknet-core` should be negligible. This might have to do with some cross-lang LTO settings. Don't have time to investigate though.

PR #335, opened 2023-03-28
fix: sierra artifact schema

tarrencev:


xJonathanLEI:
Hi thanks for the PR! It's a bad idea that I added the ERC20 Cairo 1 artifacts directly. I should have added source files instead and make a script for generating those, like how Cairo 0 artifacts are generated right now. Otherwise there's no proper way to update those when Cairo version changes.

I'm working on that separately, and will come back to this PR once that lands so that this PR only needs to deal with the Rust code changes.

xJonathanLEI:
I just checked the Cairo `1.0.0-alpha.6` code and it looks like they also added an `enum` variant on top of the new `struct` you're implementing here.

tarrencev:
Hm yeah. The CASM parsers is broken too, the renamed `hints` to `pythonic_hints` and introduced a new hint dsl

xJonathanLEI:
Yeah the casm part needs some non-trivial rework since they also changed the hash calculation (in a good way though; I actually asked them to change it lol). Will limit the scope of this PR to Sierra only then.

tarrencev:
Ah that is great to know since we are using the hash calculation now in dojo.

Also, `cairo-lang-starknet` provides a `CasmContractClass::from_contract_class(contract_class, true)` which is quite useful for taking sierra artifacts and converting them to the casm contract class. Converting from `CasmContractClass` to the starknet-rs `CompiledClass` to compute the hash was a bit tricky. I ended up just serializing to json and then deserializing with starknet-rs. Might be useful to better support going from compiler sierra artifacts to casm in starknet-rs at some point

xJonathanLEI:
> Might be useful to better support going from compiler sierra artifacts to casm in starknet-rs at some point

Honestly not sure if it's a good idea to make an SDK statically link to the compiler. Would likely blow up in binary size.

tarrencev:
> > Might be useful to better support going from compiler sierra artifacts to casm in starknet-rs at some point
> 
> Honestly not sure if it's a good idea to make an SDK statically link to the compiler. Would likely blow up in binary size.

Yeah I agree

xJonathanLEI:
Hope you don't mind that I basically reworked the whole PR lol.

PR #339, opened 2023-03-30
feat: compiled class hash calculation

xJonathanLEI:
This PR implements (again) the compiled casm class hash calculation.

Great thanks to @ArielElp for taking the feedback and changing the algo! It's much much much cleaner and easier to implement now.


xJonathanLEI:
@tarrencev FYI this should now give you the correct casm hash.

Issue #313, opened 2023-02-25
Add support for Starknet v0.11.0


xJonathanLEI:
New Starknet version [dropped](https://github.com/starkware-libs/cairo-lang/releases/tag/v0.11.0-pre). Need to add support.

- [x] support Poseidon hash (already done in #312)
- [x] support parsing Cairo 1 contract artifacts
- [x] support computing class hashes of Cairo 1 contracts
- [x] support sending Declare v2 transactions
- [x] resolve sequencer gateway [API changes](https://docs.starknet.io/documentation/starknet_versions/upcoming_versions/#api_changes)
- [x] add support for new sequencer gateway methods

PR #343, opened 2023-03-31
feat: support declare v2 transaction

xJonathanLEI:
There you go. `DECLARE` v2 supported. It's now possible to declare Cairo 1 contracts with `starknet-rs`.

However, this only partially resolves task 4 of #313:

> support sending Declare v2 transactions

as JSON-RPC support is not enabled yet. With this PR declaration is only possible via the sequencer gateway.

This is because the current specs version (0.2.1) does not support `DECLARE` v2. We will need to manually patch the spec to make it work, which is quite some work, and I will deal with that in a separate PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1568,6 +1568,7 @@ name = "starknet-accounts"
 version = "0.1.0"
 dependencies = [
  "async-trait",
+ "serde",
  "serde_json",
  "starknet-core",
  "starknet-providers",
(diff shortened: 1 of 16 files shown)


xJonathanLEI:
Note that unlike previous upgrades where I'd just ditch old transaction versions (and only keep the raw request structs), in this one I actually kept support for both `DECLARE` v1 and v2 with the high-level account constructs, since unlike previous transaction version iterations, `DECLARE` v2 does not completely replace v1 - Cairo 0 contracts can only be declared with `DECLARE` v1.

See the [example](https://github.com/xJonathanLEI/starknet-rs/blob/a5a51c858ceb4146c42bf1cc5fb643c9b6093ee9/examples/declare_cairo0_contract.rs) for declaring legacy Cairo 0 contracts and [this one](https://github.com/xJonathanLEI/starknet-rs/blob/a5a51c858ceb4146c42bf1cc5fb643c9b6093ee9/examples/declare_cairo1_contract.rs) for Cairo 1 contracts.

PR #355, opened 2023-04-18
perf: projective points for verification of ECDSA

Oppen:
Multiplication of EC points by scalars was the major bottleneck in that function. Converting the points to projective points first leads to 85% less time spent on the function according to Criterion.

### DIFF
--- a/starknet-crypto/src/ecdsa.rs
+++ b/starknet-crypto/src/ecdsa.rs
@@ -1,6 +1,6 @@
 use starknet_curve::{
     curve_params::{EC_ORDER, GENERATOR},
-    AffinePoint,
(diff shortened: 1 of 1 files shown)


xJonathanLEI:
@Oppen Thanks for the PR but looks like there's a clippy error.

xJonathanLEI (approved):
LGTM. Thanks for the PR!

I just checked on my machine. Before:

    ecdsa_verify            time:   [3.1499 ms 3.1645 ms 3.1799 ms]

and after:

    ecdsa_verify            time:   [436.92 µs 439.40 µs 442.01 µs]

An 85% improvement indeed!

PR #377, opened 2023-05-11
feat(jsonrpc): server-side request parsing

xJonathanLEI:
This PR adds a new public type `JsonRpcRequest` for easy parsing of jsonrpc requests on the server side. A new example `parse_jsonrpc_request` has also been added to demonstrate how it would work. With this new type, jsonrpc server implementations would simply need to write handlers for the different data variants - no need to fiddle with parsing params or mapping request data types to methods names.

Do note that the automatic generation of response types hasn't been implemented yet, meaning server implementations would still need to manually put together the response types at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,8 @@ Examples can be found in the [examples folder](./examples):
 
 8. [Deploy an Argent X account to a pre-funded address](./examples/deploy_argent_account.rs)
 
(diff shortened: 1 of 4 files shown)


xJonathanLEI:
After the response type generation is done, I will add an example to demonstrate _really_ running a simple HTTP server for it. For now it's just parsing.

PR #410, opened 2023-06-23
feat: udc deployed address calculation

xJonathanLEI:
Based on the [UDC implementation](https://github.com/OpenZeppelin/cairo-contracts/blob/70cbd05ed24ccd147f24b18c638dbd6e7fea88bb/src/openzeppelin/utils/presets/UniversalDeployer.cairo).


milancermak:
This would be a sweet feature in starkli... just sayin'

xJonathanLEI:
@milancermak That's exactly why this is added :))

Issue #400, opened 2023-06-13
starknet contract 2.0 support


moodysalem:
the compilation artifact of a 2.0 contract looks different in the abi key:
- all the functions seem to be contained in a new interface type
- there is a constructor type
- there is a new impl type

xJonathanLEI:
2.x contract artifacts haven't been officially released yet but since release candidates are available now I can take a look into this. Thanks for spotting this!

xJonathanLEI:
Just checked and the changes are only on the ABI - no structural changes otherwise.

I've finished supporting the new ABI types on the `dev/cairo_2_abi` branch. I will add test cases and merge and release once `2.0.0` is officially out.

xJonathanLEI:
Just force pushed the `dev/cairo_2_abi` branch with:

1. Test artifacts generated from `v2.0.0-rc4`;
2. A fix for getting different Sierra class hashes as `cairo-lang`.

The fix is not strictly needed: for Cairo 1 declarations the ABI is just a flattened string anyways, so having different serialized ABI compared to `cairo-lang` wouldn't fail declarations. However, to avoid confusion, I'm still aligning the behavior with the "official" hash.

PR #413, opened 2023-06-23
fix: default event inputs

tarrencev:
cairo outputs abis with a default event with no inputs:

https://github.com/starkware-libs/cairo/blob/d80d29b52f8972f63ceb490567b26172245e95ff/crates/cairo-lang-starknet/test_data/minimal_contract.sierra.json#L468


xJonathanLEI:
See #400.

Issue #403, opened 2023-06-16
`no_std` support for `starknet-core`


xJonathanLEI:
Tracking issue for making the `starknet-core` crate `no_std`-compatible.

For reference, #306 made `starknet-crypto` `no_std`-compatible, and #308 added CI workflows for testing the compatibility.

dbejarano820:
I can tackle

dbejarano820:
Just as an update, this is a WIP

exp-table:
LFG

PR #415, opened 2023-06-25
no-std support for starknet-core

dbejarano820:
We want to add no-std support for the `starknet-core` crate

#403

### DIFF
--- a/starknet-core/src/lib.rs
+++ b/starknet-core/src/lib.rs
@@ -1,6 +1,19 @@
+#![cfg_attr(not(feature = "std"), no_std)]
 #![allow(clippy::comparison_chain)]
 #![doc = include_str!("../README.md")]
 
+#[cfg(feature = "std")]
+include!("./with_std.rs");
+
+#[cfg(not(feature = "std"))]
+include!("./without_std.rs");
+
+pub mod stdlib {
+    #[cfg(feature = "std")]
+    pub use crate::with_std::*;
+    #[cfg(not(feature = "std"))]
+    pub use crate::without_std::*;
+}
 pub mod serde;
 
 pub mod types;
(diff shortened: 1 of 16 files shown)


xJonathanLEI:
Thanks for working on this, but it seems that we're losing lot of functionality in `no_std` mode?

dbejarano820:
> Thanks for working on this, but it seems that we're losing lot of functionality in `no_std` mode?

Mainly in the legacy contract, if I'm not mistaken there's no way around not using std::io in no-std . @tdelabro can correct me if I'm wrong

xJonathanLEI:
Not just legacy contract right? Looks like you can't do rather basic stuff like computing Sierra class hashes either.

starknet-core/src/lib.rs
```diff
@@ -1,6 +1,19 @@
...
+
+#[cfg(not(feature = "std"))]
+include!("./without_std.rs");
+
+pub mod stdlib {
+    #[cfg(feature = "std")]
+    pub use crate::with_std::*;
+    #[cfg(not(feature = "std"))]
+    pub use crate::without_std::*;
+}
```
tdelabro, line 16: As the rest of the repo doesn't use this method to support no_std, I would suggest continuing doing it the same way: at the file level, by importing from core and alloc rather than std + have hashbrown as a dependency

tdelabro (changes requested):

dbejarano820:
> @xJonathanLEI the problem is that serde_json has a really poor support for json serialization when in `no_std`. Everything iteresting is not exposed.
> 
> 
> 
> And there is this weird thing where in order to have a contract hash, we compute the hash of the python `json.dump` method. Do you know if this will be changed in the future?
> 
> 
> 
> Otherwise the only way out of this I see is a manual implementation of the `to_string_pythonic` method that doesn't uses `serde_json` 

@xJonathanLEI Lmk the best approach we should take here and i'll include everything in my next commit

xJonathanLEI:
Superseded by #428.

PR #421, opened 2023-07-01
fix: missing L1Handler abi type

glihm:
Same as a function, only the name is changing.


xJonathanLEI:
The ERC20 contract is not supposed to be changed. It's copied directly from the `cairo` repo.

glihm:
> The ERC20 contract is not supposed to be changed. It's copied directly from the `cairo` repo.

Arg sorry, then we should do an other test contract?
Thanks for the feedback. 👍🏻

PR #438, opened 2023-07-07
Remove last digit in sha2 version, allowing users to bump version, or use the one they need

MauroToscano:
Pinning last digit in version forces the users of the library to use that version. Else unsolved conflicts may appear.

This also makes dependencies outdated. Right now for example there's a new version of sha2.

By omitting the last digit, it uses the last one by default, or the one specified by the user of the library.

For the record, this is the latest version of sha2 right now: https://github.com/RustCrypto/hashes/commit/d8b088aece5659cecdce397ce48d6129b6c4c440


MauroToscano:
This shouldn't be needed for upgraded versions

PR #463, opened 2023-08-17
test: make hardcoded RPC address configurable

### LINKED ISSUE #462: Make hardcoded RPC address configurable
We want to incorporate starknet-rs test in our CI.
Making the rpc address to be overridable by an env variable would be really helpful.

https://github.com/NethermindEth/juno/issues/1114

omerfirmak:
Fixes #462

### DIFF
--- a/starknet-accounts/tests/single_owner_account.rs
+++ b/starknet-accounts/tests/single_owner_account.rs
@@ -16,7 +16,6 @@ use starknet_providers::{
     Provider, SequencerGatewayProvider,
 };
 use starknet_signers::{LocalWallet, SigningKey};
-use std::env;
 use std::sync::Arc;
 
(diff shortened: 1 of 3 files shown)


xJonathanLEI:
CI red. Mind fixing?

omerfirmak:
Fixed formatting, I think some tests are flakey tho.

xJonathanLEI:
Don't worry about that part. It seems to be a pathfinder issue. I will retry if they fail.

BTW some tests are marked as `ignore` due to pathfinder bugs (fixed in 0.7.2 but I'm still running 0.7.1 as of now). If you want to run all tests you need to do something like `cargo test --all -- --include-ignored`.

omerfirmak:
> `cargo test --all -- --include-ignored`

Nice, good to know. Thanks!

PR #494, opened 2023-11-01
fix: execution status revert reason field is optional

fracek:
The Starknet JSON-RPC spec defines this field as optional. This commit fixes compatibility with nodes like Juno that don't include this field if the revert reason is empty.

### DIFF
--- a/starknet-core/src/types/execution_result.rs
+++ b/starknet-core/src/types/execution_result.rs
@@ -61,17 +61,19 @@ impl<'de> Deserialize<'de> for ExecutionResult {
         #[cfg_attr(feature = "no_unknown_fields", serde(deny_unknown_fields))]
         struct Raw {
             execution_status: TransactionExecutionStatus,
-            #[serde(default)]
-            revert_reason: String,
+            revert_reason: Option<String>,
         }
 
         let raw = Raw::deserialize(deserializer)?;
 
-        match raw.execution_status {
-            TransactionExecutionStatus::Succeeded => Ok(Self::Succeeded),
-            TransactionExecutionStatus::Reverted => Ok(Self::Reverted {
-                reason: raw.revert_reason,
+        match (raw.execution_status, raw.revert_reason) {
+            (TransactionExecutionStatus::Succeeded, None) => Ok(Self::Succeeded),
+            (TransactionExecutionStatus::Reverted, reason) => Ok(Self::Reverted {
+                reason: reason.unwrap_or_default(),
             }),
+            (TransactionExecutionStatus::Succeeded, Some(_)) => Err(serde::de::Error::custom(
+                "field `revert_reason` must not exist when `execution_status` is `SUCCEEDED`",
+            )),
         }
     }
 }
--- a/starknet-core/src/types/execution_result.rs
+++ b/starknet-core/src/types/execution_result.rs
@@ -61,20 +61,17 @@ impl<'de> Deserialize<'de> for ExecutionResult {
         #[cfg_attr(feature = "no_unknown_fields", serde(deny_unknown_fields))]
         struct Raw {
             execution_status: TransactionExecutionStatus,
-            revert_reason: Option<String>,
+            #[serde(default)]
+            revert_reason: String,
         }
 
         let raw = Raw::deserialize(deserializer)?;
 
-        match (raw.execution_status, raw.revert_reason) {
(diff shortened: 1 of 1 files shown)


xJonathanLEI:
While the spec does say it's optional, I figured that it simply meant that it's `null` when the transaction succeeds? Is there a legitimate case where the revert reason is empty when the transaction reverts?

xJonathanLEI:
But in any case I like this PR adds compatability without breaking changes on the API surface. However, one part I'd like to see changed is how the new implementation no longer rejects the combination of a successful execution with revert reason - I believe this apparently an invalid combination.

fracek:
> While the spec does say it's optional, I figured that it simply meant that it's `null` when the transaction succeeds? Is there a legitimate case where the revert reason is empty when the transaction reverts?

Juno (both in RPC 0.4 and 0.5) doesn't return the field if the reason is empty. You can see that making the following call on mainnet.

```json
{
	"id": 0,
	"jsonrpc": "2.0",
	"method": "starknet_getTransactionReceipt",
	"params": ["0x0527505db95eca337e3ad224ff3881e8565403f817fcbef1a2364c3ed054d11c"]
}
```

<details>
  <summary>RPC 0.4</summary>

```json
{
	"jsonrpc": "2.0",
	"result": {
		"type": "INVOKE",
		"transaction_hash": "0x527505db95eca337e3ad224ff3881e8565403f817fcbef1a2364c3ed054d11c",
		"actual_fee": "0x24143b000377",
		"execution_status": "REVERTED",
		"finality_status": "ACCEPTED_ON_L1",
		"block_hash": "0x405453ad0c33bd9515cfb39e268a247b908cc16c996abc99f5955e9143204e2",
		"block_number": 367818,
		"messages_sent": [],
		"events": [
			{
				"from_address": "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
				"keys": [
					"0x99cd8bde557814842a3121e8ddfd433a539b8c9f14bf31ebf108d12e6196e9"
				],
				"data": [
					"0x6f5ccf18f9a80aca492ea25c50f9916fdd1db207d09c7a1bd539b3c1303ce12",
					"0x1176a1bd84444c89232ec27754698e5d2e7e1a7f1539f12027f28b23ec9f3d8",
					"0x24143b000377",
					"0x0"
				]
			}
		]
	},
	"id": 0
}
```
</details>

<details>
  <summary>RPC 0.5</summary>

```json
{
	"jsonrpc": "2.0",
	"result": {
		"type": "INVOKE",
		"transaction_hash": "0x527505db95eca337e3ad224ff3881e8565403f817fcbef1a2364c3ed054d11c",
		"actual_fee": "0x24143b000377",
		"execution_status": "REVERTED",
		"finality_status": "ACCEPTED_ON_L1",
		"block_hash": "0x405453ad0c33bd9515cfb39e268a247b908cc16c996abc99f5955e9143204e2",
		"block_number": 367818,
		"messages_sent": [],
		"events": [
			{
				"from_address": "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
				"keys": [
					"0x99cd8bde557814842a3121e8ddfd433a539b8c9f14bf31ebf108d12e6196e9"
				],
				"data": [
					"0x6f5ccf18f9a80aca492ea25c50f9916fdd1db207d09c7a1bd539b3c1303ce12",
					"0x1176a1bd84444c89232ec27754698e5d2e7e1a7f1539f12027f28b23ec9f3d8",
					"0x24143b000377",
					"0x0"
				]
			}
		],
		"execution_resources": {
			"steps": "0x0",
			"memory_holes": "0x0",
			"pedersen_builtin_applications": "0x0",
			"range_check_builtin_applications": "0x0",
			"bitwise_builtin_applications": "0x0",
			"ecdsa_builtin_applications": "0x0",
			"ec_op_builtin_applications": "0x0",
			"keccak_builtin_applications": "0x0",
			"poseidon_builtin_applications": "0x0"
		}
	},
	"id": 0
}
```
</details>

> But in any case I like this PR adds compatability without breaking changes on the API surface. However, one part I'd like to see changed is how the new implementation no longer rejects the combination of a successful execution with revert reason - I believe this apparently an invalid combination.

True. I'm concerned that at some point in the future there will be a client that sends a successful status with an empty error message (because why not?) since it's technically following the spec. Happy to change the code if you think the risk is too low.

xJonathanLEI (approved):
LGTM

PR #501, opened 2023-11-08
feat: add support for `flat` event kind

### LINKED ISSUE #499: Add support for `flat` event
Cairo has a new [flat event](https://github.com/starkware-libs/cairo/pull/4112). The rationale is [described here](community.starknet.io/t/cairo-v2-3-0-release-candidate/101137).

In the code it looks [something like this](https://github.com/OpenZeppelin/cairo-contracts/blob/4eafccf107f98d975b20cbb6544b2a998e5f403f/src/tests/mocks/ownable_mocks.cairo#L24).

In the ABI, it's something like this (note this example is from a different contract):

```json
{
            "type": "event",
            "name": "twostep::one::one::Event",
            "kind": "enum",
            "variants": [
                {
                    "name": "OwnableEvent",
                    "type": "openzeppelin::access::ownable::ownable::OwnableComponent::Event",
                    "kind": "flat"
                },
                {
                    "name": "UpgradeableEvent",
                    "type": "openzeppelin::upgrades::upgradeable::UpgradeableComponent::Event",
                    "kind": "flat"
                }
            ]
        }
```

This isn't currently supported by the lib.

Found out while trying to declare a contract with a flat event using starkli, which failed with `Error: failed to parse contract artifact`.

xJonathanLEI:
Fixes #499.

This is technically a breaking change, as we're adding a enum variant to `EventFieldKind`, but it would only break applications that exhaust all variants in a `match` expression, which I deem rather unlikely. If we want to make a release on top of `0.7.2` for this, I might just apply romantic versioning an make it a patch release as `0.7.3` instead of bumping minor to `0.8.0`.


milancermak:
No one is going to fault you for 0.7.3 I imagine, especially with the wild west versioning going on in the ecosystem 😁

PR #505, opened 2023-11-23
feat(jsonrpc): switch to spec 0.5.1

xJonathanLEI:
Here we go 0.5.x.


glihm:
Could be great to have your feedback on #483 to be compliant with the JSON RPC spec, now that the starknet spec is using the `data` field on the error at least once in `CONTRACT_ERROR`.

Also this will be very helpful to have the `data` field returned by providers, important for debugging in some situations.

PR #506, opened 2023-11-26
feat(starknet-core): initialize `EthAddress` from bytes

glihm:
I am currently working with Ethereum data, and I didn't find an easy way to create an `EthAddress` from a buffer of bytes.

This PR aims at adding implementations to create an `EthAddress` from an array and a slice.

Please don't hesitate is there is already an easy to do so that I missed, I'll close the PR in that case.

Thanks for the consideration.


starknet-core/src/types/eth_address.rs
```diff
@@ -31,12 +31,18 @@ mod errors {
     #[derive(Debug)]
     pub struct FromFieldElementError;
 
+    #[derive(Debug)]
+    pub struct FromBytesSliceError(pub alloc::string::String);
+
```
xJonathanLEI, line 36: I don't think we need a `String` for this error.

starknet-core/src/types/eth_address.rs
```diff
@@ -156,3 +168,75 @@ impl From<EthAddress> for FieldElement {
...
+                "expected 20 bytes found {}",
+                value.len()
+            )));
+        }
+
+        Ok(Self {
+            inner: value
+                .try_into()
+                .map_err(|e| FromBytesSliceError(format!("can't convert slice to array: {}", e)))?,
+        })
```
xJonathanLEI, line 187: No error mapping is needed here I think? I believe the only reason this can fail here is a length mismatch, which we already checked for.

xJonathanLEI (changes requested):

starknet-core/src/types/eth_address.rs
glihm: That's a very good point, will edit accordingly.

starknet-core/src/types/eth_address.rs
glihm: Removed.

starknet-core/src/types/eth_address.rs
glihm: Changed by unwrap with a comment on the unwrap safety.

xJonathanLEI (approved):

PR #510, opened 2023-12-02
feat(sequencer): deprecated endpoint error code

### LINKED ISSUE #509: Bug: There are two bugs in the RPC error handling.
```shell
[2023-12-02T06:05:37Z TRACE starknet_providers::jsonrpc::transports::http] Response from JSON-RPC: {"jsonrpc":"2.0","error":{"code":-32603,"message":"Internal error","data":{"error":"Internal error: Error in the called contract (0x05686c52b6f38639eb9cfb3dfff1b3260315099aa045fcc0b4a865068ba36aad):\nError at pc=0:558:\nCouldn't compute operand op1. Unknown value for memory cell 7:5"}},"id":1}

called `Result::unwrap()` on an `Err` value: Provider(Other(TransportError(Json(Error("data did not match any variant of untagged enum JsonRpcResponse", line: 0, column: 0)))))

```

clint419:
Closes #509 
@xJonathanLEI

### DIFF
--- a/starknet-providers/src/sequencer/mod.rs
+++ b/starknet-providers/src/sequencer/mod.rs
@@ -95,6 +95,8 @@ pub enum ErrorCode {
     DuplicatedTransaction,
     #[serde(rename = "StarknetErrorCode.INVALID_CONTRACT_CLASS")]
     InvalidContractClass,
+    #[serde(rename = "StarknetErrorCode.DEPRECATED_ENDPOINT")]
+    DeprecatedEndpoint,
 }
 
 impl SequencerGatewayProvider {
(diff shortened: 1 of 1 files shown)


xJonathanLEI:
I see you're adding 2 things:

1. support for parsing the `data` field when there's an error
2. support for the new deprecated endpoint error code from the sequencer

1 has been implemented by #513 (in a more idiomatic way too). I will now rebase this PR to only include 2. Thanks for the fix.

xJonathanLEI:
In fact, the best way to deal with the deprecated endpoints would be to simply remove them for good, but let's leave that for another day.

PR #522, opened 2023-12-10
feat: add byte array type and string conversions

glihm:
This PR aims at bringing support for [ByteArray](https://github.com/starkware-libs/cairo/blob/a4de08fbd75fa1d58c69d054d6b3d99aaf318f90/corelib/src/byte_array.cairo) type.

### DIFF
--- a/starknet-core/src/types/byte_array.rs
+++ b/starknet-core/src/types/byte_array.rs
@@ -1,0 +1,348 @@
+//! Support for string compatibility with Cairo `ByteArray`.
+//! https://github.com/starkware-libs/cairo/blob/a4de08fbd75fa1d58c69d054d6b3d99aaf318f90/corelib/src/byte_array.cairo
+//!
+//! The basic concept of this `ByteArray` is relying on a string being
+//! represented as an array of bytes packed by 31 bytes in a felt.
+//! To support any string even if the length is not a multiple of 31,
+//! the `ByteArray` struct has a `pending_word` field, which is the last
+//! word that is always shorter than 31 bytes.
+//!
+//! In the data structure, everything is represented as a felt to be compatible
+//! with the Cairo implementation.
+use alloc::string::FromUtf8Error;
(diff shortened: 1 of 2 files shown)


glihm:
I think I've JSON rpc related errors, if you have the opportunity to manually trigger the workflow it could be great @xJonathanLEI. Thanks a lot!

glihm:
I'm thinking that we should introduce a Bytes31 type to ensure that the `data` is actually with the correct format. WDYT @xJonathanLEI?

glihm:
@xJonathanLEI here's the updated version of the PR. Few months of blank here, sorry for that.

I've added the `Bytes31` type to be consistent with Cairo.

Happy to address any comment!

glihm:
@xJonathanLEI rebased on master, happy to dive in any update as necessary.

xJonathanLEI:
Hey @glihm thanks for the PR. I see you're modelling the types here directly after their Cairo implementations, with the internals exposed.

It kinda makes me wonder though: do we really expect users to care about what a particular array's `data`, `pending_word`, and `pending_word_len` really are?

I'm not exactly convinced that it's the case. The exact composition should only matter when crossing the Rust <-> Cairo boundary (i.e. encoding/decoding), which should be mostly opaque to users. Users most likely just want to deal with bytes.

So IMO a better design would be a fully encapsulated `ByteArray` type that's internally backed by `Vec<u8>`. We _could_ provide getters for viewing the components like `data`, `pending_word`, and `pending_word_len`, but I also don't think that's actually needed.

Then we should have this type implement `Encode` and `Decode` to make crossing the Cairo boundary seamless.

xJonathanLEI:
Superseded by #682.

elton-cs:
No way! I was just looking for a ByteArray implementation and turns out it was implemented 5 days ago! Man, we really are on the bleeding edge haha 😅. Thanks for the new feature guys!

PR #523, opened 2023-12-11
feat(jsonrpc): switch to spec 0.6.0

xJonathanLEI:
Moving away from JSON-RPC v0.5.1 to v0.6.0. Applications really should use v0.6.0 unless they have a very good reason to stay on v0.5.1.


xJonathanLEI:
Switched to Juno v0.8.0 for the integration tests as unfortunately Pathfinder v0.10.1 has a number of issues with its v0.6.0 implementation. Testing with Juno is more flaky though, as it doesn't implement automatic retry on rate limit errors. Will switch back to Pathfinder once the issues are resolved with a new release.

PR #527, opened 2023-12-13
refactor: use `deserialize_str` instead

### LINKED ISSUE #526: `FieldElement` doesn't seem to support deserializing from non self-describing format
https://github.com/xJonathanLEI/starknet-rs/blob/0df9ad3417a5f10d486348737fe75659ca4bcfdc/starknet-ff/src/lib.rs#L577-L584

Not sure if this is intended, but the use of `deserialize_any` in `FieldElement`'s `Deserialize` impl means I cant use bincode/postcard with it as they don't support the method. Deserializing `FieldElement` from bincode/postcard will result in an error like:

```
Bincode does not support the serde::Deserializer::deserialize_any method
```

And as mentioned in serde [website](https://serde.rs/impl-deserialize.html), 

> When implementing Deserialize, you should avoid relying on Deserializer::deserialize_any unless you need to be told by the Deserializer what type is in the input. Know that relying on Deserializer::deserialize_any means your data type will be able to deserialize from self-describing formats only, ruling out Postcard and many others.

kariy:
Resolves #526

### DIFF
--- a/starknet-ff/src/lib.rs
+++ b/starknet-ff/src/lib.rs
@@ -579,7 +579,7 @@ mod serde_field_element {
         where
             D: serde::Deserializer<'de>,
         {
-            deserializer.deserialize_any(FieldElementVisitor)
+            deserializer.deserialize_str(FieldElementVisitor)
         }
     }
 


xJonathanLEI:
Thanks! Just curious: for bincode though, wouldn't you want to use bytes instead of str for the repr? AFAIK `serde` doesn't play well with multiple formats.

On the serialization side, the only option is to use newtypes (e.g. `struct SerializeAsBytes(FieldElement)`).

On the deserialization side, either use the newtype option, or just go with `deserialize_any` and implement many `visit_*` (but then you run into the issue this PR tries to solve).

xJonathanLEI:
Or, of couse, you can always mark all your `FieldElement` occurrences with `#[serde(with = "...")]`, or even go as far as manual impls. I guess what the "official" `Serialize`/`Deserialize` impls here only serve as the default. Maybe `starknet-ff` itself can offer alternative modules to be used with `#[serde(with = "...")]`, if that's helpful.

kariy:
> Thanks! Just curious: for bincode though, wouldn't you want to use bytes instead of str for the repr?

Ideally yes

> Or, of couse, you can always mark all your FieldElement occurrences with #[serde(with = "...")], or even go as far as manual impls.

I could do this but i dont quite like having to tag `#[serde(..)]` on every types that use it or even a newtype struct.

> I guess what the "official" Serialize/Deserialize impls here only serve as the default. Maybe starknet-ff itself can offer alternative modules to be used with #[serde(with = "...")], if that's helpful.

i just found out about [`Deserializer::is_human_readable()`](https://docs.rs/serde/latest/serde/de/trait.Deserializer.html#method.is_human_readable)  . Maybe we can conditionally call the preferred deserialize entry points like this ?

```rust
 if deserializer.is_human_readable() {
            deserializer.deserialize_any(FieldElementVisitor)
        } else {
            deserializer.deserialize_bytes(FieldElementVisitor)
        }
}        
```

deserializer like `serde_json` would return `true` while something like bincode would return `false`, so it would still maintain the same behaviour as before but should now work with non self-describing formats.

this is also what `alloy` is [using](https://github.com/alloy-rs/core/blob/c2f6ad2993209a94047ff395f1364b4ba3c65bf3/crates/primitives/src/bits/serde.rs#L61) for their primitive types

xJonathanLEI:
Thanks! I didn't know that.

This is sorta a breaking change though. Non-human-readable self-describing formats used to be serialized as strings but now they use bytes. Files previously saves with such a format will become unreadable.

I guess we can do this with a breaking semver bump.

kariy:
I'm happy to contribute on that if you're okay with the approach.

xJonathanLEI:
Yeah that's certainly a good approach. We just need to bump a breaking version with this change.

xJonathanLEI:
@kariy [This from `rust_decimal`](https://github.com/paupino/rust-decimal/blob/829355a377ef01affbd53a0c12de431c6904173e/src/serde.rs#L279-L298) is an interesting approach too. One issue with this approach though: someone enabling this feature in the dependency tree would affect everyone else.

PR #530, opened 2023-12-18
fix: skip serialization of continuation token if `None`

apoorvsadana:
In the Starknet Spec, the `continuation_token` is not a required field. So in the last event page, the `continuation_token` must be absent from the response instead of being null. I also tried this with the infura RPC (which uses pathfinder?) and works in the same way. We are integrating Madara with the explorer and that's breaking because of this.

### DIFF
--- a/starknet-core/src/types/mod.rs
+++ b/starknet-core/src/types/mod.rs
@@ -106,7 +106,8 @@ pub struct EventsPage {
     /// Matching events
     pub events: Vec<EmittedEvent>,
     /// A pointer to the last element of the delivered page, use this token in a subsequent query to
-    /// obtain the next page
+    /// obtain the next page. If the value is `None`, don't add it to the response as clients might
+    /// use `contains_key` as a check for the last page.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub continuation_token: Option<String>,
 }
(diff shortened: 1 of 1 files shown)


glihm (commented):
Looks good, I would only add in the comment the reason why it's added. :+1:

apoorvsadana:
Done!

xJonathanLEI:
Does this need to be backported or are you guys already using the latest commit?

apoorvsadana:
We're already using a specific `rev`. We can change that to the latest one, shouldn't be an issue.

xJonathanLEI:
Are you using an old `rev` with JSON-RPC pre-0.6.0 though? The latest commit uses pre-0.6.0.

xJonathanLEI:
Or are you only using this one single type? Then it shouldn't matter. Not sure how the lib is used in Madara.

xJonathanLEI (approved):
LGTM. Thanks for spotting. Really all `Option` fields should be skipped for serialization when it's `None`.

apoorvsadana:
We are using it for other things as well and currently Madara isn't on 0.6.0. So you're right, just updating the commit might break. We are on this `rev` right now - a35ce22. Is it possible to backport it?

xJonathanLEI:
Oh that's a very old commit, not even JSON-RPC v0.5.1.. still on v0.4.0.

xJonathanLEI:
In this case maybe you can backport this in a fork and use the fork for Madara until it upgrades?

apoorvsadana:
Ya sure we can do that then. Thanks for the help!

PR #532, opened 2023-12-21
feat: de/serialize based on human readability

kariy:
This PR improves the serde impls used to mainly de/serialize the `FieldElement` type. The de/serialization implementations now take into consideration whether `Serialize`/`Deserialize` implementations should expect to de/serialize in human-readable form or not.

This possibly(?) introduces a breaking change due to this PR #527.

### DIFF
--- a/starknet-core/src/serde/num_hex.rs
+++ b/starknet-core/src/serde/num_hex.rs
@@ -1,5 +1,6 @@
 pub mod u64 {
     use alloc::{fmt::Formatter, format};
+    use core::mem;
 
     use serde::{de::Visitor, Deserializer, Serializer};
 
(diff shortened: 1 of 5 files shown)


xJonathanLEI:
This PR is all good by itself. I wanted to merge it, but it this is an unfortunately timing to introduce a breaking change as we're migrating to the [official `Felt` type](https://github.com/starknet-io/types-rs). Asking users to deal with a breaking change only to immediately break them again via the migration doesn't sound like a good idea. That's the reason for the long delay.

kariy:
> This PR is all good by itself. I wanted to merge it, but it this is an unfortunately timing to introduce a breaking change as we're migrating to the [official `Felt` type](https://github.com/starknet-io/types-rs). Asking users to deal with a breaking change only to immediately break them again via the migration doesn't sound like a good idea. That's the reason for the long delay.

Understandable! It's not urgent for now so I don't mind the delay 👍.

xJonathanLEI:
Rebased and added a new bound check for deserializing as the new `Felt` type accepts _any_ integer input, which makes sense as they can always just take the modulo, but it makes little sense to me for deserialization purposes - you'd never serialize the value into an out-of-range value in the first place.

xJonathanLEI:
I feel like we need more tests as new functionality has been added with this one.

xJonathanLEI:
Hmm an issue with this PR is that it leaves `UfeHexOption` unchanged. So it gets a bit weird.

xJonathanLEI:
Just pushed a commit adding binary serde to `UfeHexOption` and `UfePendingBlockHash` too. When the value is `None` it encodes an empty slice.

starknet-core/src/serde/num_hex.rs
```diff
@@ -9,21 +10,29 @@ pub mod u64 {
     where
         S: Serializer,
     {
-        serializer.serialize_str(&format!("{value:#x}"))
+        if serializer.is_human_readable() {
+            serializer.serialize_str(&format!("{value:#x}"))
+        } else {
+            serializer.serialize_bytes(&value.to_be_bytes())
```
xJonathanLEI, line 16: This is actually wasteful. `serde` does not support fixed size array but we know this is actually always 8 bytes. Using a byte slice here forces the serializer to add a length prefix unnecessarily.

xJonathanLEI: Changed to serialize into `u64` directly instead.

xJonathanLEI (approved):

PR #539, opened 2024-02-01
feat(jsonrpc): make the client and transport implement `Clone`

pierre-l:


xJonathanLEI:
Sorry but PRs without context or descriptions won't be reviewed. Please give info on why you're making the changes.

PR #542, opened 2024-02-14
Fix `from_affine_point` infinity definition

tcoratger:
At the moment, to construct a projective point from an affine point, the `from_affine_point` always consider `infinity` as false. 

However, this will be wrong if the affine point is the identity point, for which `infinity` has to be `true`. 

For this purpose, the `from_affine_point` method is modified to always take into account the `infinity` attribute from the provided `p` affine point during the construction of the projective point.


xJonathanLEI (approved):
LGTM!

PR #543, opened 2024-02-15
fix `add_assign` for `ProjectivePoint`

tcoratger:
At the moment, when using `add_assign` for `ProjectivePoint`, the case P + (-P) was not taken into account, leading to 2P instead of infinity point. 

For this purpose a piece of logic is added to check, during P1 + P2 addition and in the case where x1 = x2, if y1 = y2 (should return point doubling) or if y1 = -y2 (should return infinity point).

Also some of the logic on other implementations of `add_assign` is simplified to remove some useless code.


xJonathanLEI (approved):
LGTM. Thanks! Will make a patch release for `starknet-curve` right away.

PR #546, opened 2024-02-15
feat: implement `Neg` for `AffinePoint`

tcoratger:
## Description
This pull request includes the following changes:
- Rearranged imports in all files to ensure proper classification in alphabetical order.
- Added `ops::Neg` implementation for `AffinePoint`, improving usability during subtraction operations.
- Removed Clippy warning on `ops::Mul` implementation for `AffinePoint` by enhancing the implementation.

### DIFF
--- a/starknet-accounts/src/account/declaration.rs
+++ b/starknet-accounts/src/account/declaration.rs
@@ -3,7 +3,6 @@ use super::{
     LegacyDeclaration, PreparedDeclaration, PreparedLegacyDeclaration, RawDeclaration,
     RawLegacyDeclaration,
 };
-
 use starknet_core::{
     crypto::compute_hash_on_elements,
     types::{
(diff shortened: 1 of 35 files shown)


xJonathanLEI (approved):
Pushed a commit to undo some of the changes:

- import list changes: I believe it's a common practice to divide `use` statements into 3 groups: `std`/`core`/`alloc`; then external crates; then internal. That currently isn't exactly upheld in the codebase, but I also don't want to completely change it into another style. Also a file `codegen.rs` that's not supposed to be manually edited has been edited in the PR.
- replacing `for` with `.fold()`: I've run benchmark on my laptop and this change slows all EC operations down by 2-5% on average. I guess it's not a good idea to use fancy `iter` methods in hot code paths.

Otherwise LGTM.

xJonathanLEI:
Interesting. But it looks like that same change can be applied to the `for` impl. I think in general `iter` is just slower.

tcoratger:
> Interesting. But it looks like that same change can be applied to the `for` impl. I think in general `iter` is just slower.

Yes exactly I meant while keeping the `for` there is no problem on that. I put the iter to remove the clippy warning but if it is slower then it is better to stay with the old formalism with the `for`. But by applying the change from #553 to `for` as proposed in #553, you should see performance gains especially on small numbers with lots of leading zeros.

PR #548, opened 2024-02-17
refactor curve stark

DoTheBestToGetTheBest:
Small refactor here.
Part 1 : 
Introduction of a Shared Lambda Calculation Function (calculate_lambda) : I just created a versatile function named "calculate_lambda' to centralize the calculation of the Lambda.

The function has design for two case: 

Point doubling: in tis case lambda is calclted using the formula Lambda = (3x^3 +1)/2y for a point represented by coordinates (x,y). This case is speciic to doubling op.

Point addition: When we deal with addition of two point in the curve which are (x1, y1) and (x2, y2) the corditnates of the two poing are added.

The function switches between these formulas based on the input it receives.

Part 2:
Purpose of the Identity Trait
The Identity trait was created to define a common interface for obtaining the identity element of elliptic curve points. 
The  trait provides a standardized way to access the identity element for different types of points.
 
That's all for me.

### DIFF
--- a/starknet-curve/src/ec_point.rs
+++ b/starknet-curve/src/ec_point.rs
@@ -4,6 +4,11 @@ use crate::curve_params::{ALPHA, BETA};
 
 use core::ops;
 
+/// Trait to define identity for different point types.
+pub(crate) trait Identity {
+    fn identity() -> Self;
+}
+
 /// A point on an elliptic curve over [FieldElement].
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub struct AffinePoint {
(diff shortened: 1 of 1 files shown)


starknet-curve/src/ec_point.rs
```diff
@@ -20,6 +25,49 @@ pub struct ProjectivePoint {
...
+        // Point doubling case
+        (None, None) => {
+            let dividend = FieldElement::THREE * (x1 * x1) + FieldElement::ONE;
+            let divisor_inv = y1.double().invert().expect("Inversion failed in doubling");
+            dividend * divisor_inv
+        }
+        // Point addition case
+        (Some(x2), Some(y2)) => {
+            let numerator = y2 - y1;
+            let denominator_inv = (x2 - x1).invert().expect("Inversion failed in addition");
```
tcoratger, line 45: Here I would say that there is a danger in doing this here, because previously this operation was done in the logical sequence of the function and therefore at the step where it was done, we were sure that x2 was not equal at x1 (the check was done before). So there was no risk of a null denominator. Whereas with this utility, we can take the risk, due to ignorance of the inside of the function for the user, of using it with an x1 which would be equal to an x2, generating a panic, right?

DoTheBestToGetTheBest: > Here I would say that there is a danger in doing this here, because previously this operation was done in the logical sequence of the function and therefore at the step where it was done, we were sure that x2 was not equal at x1 (the check was done before). So there was no risk of a null denominator. Whereas with this utility, we can take the risk, due to ignorance of the inside of the function for the user, of using it with an x1 which would be equal to an x2, generating a panic, right?

Thank you so much for your review!

Hmm i don't really understand how it can be 0 here

in the function where we use this, there is a check which is 


```rust
if self.x == rhs.x {
    if self.y == rhs.y {
        self.double_assign();
    } else {
        *self = AffinePoint::identity();
    }
    return;
}
```

So basically if i'm not wrong understanting this part, this check ensures that when x1 (which is self.x) and x2 (which is rhs.x) are equal, the function doesn't proceed to the lambda calculation for addition but instead handles it as either a point doubling or a point at infinit


let me know please :)
And thank you!

tcoratger: Yes in this case it's true but I mean in general circumstances, as a user I can use this function without checking before that x1 is not equal to x2 leading potentially to a panic. Or maybe we should wrap this into an error?

DoTheBestToGetTheBest: > Yes in this case it's true but I mean in general circumstances, as a user I can use this function without checking before that x1 is not equal to x2 leading potentially to a panic. Or maybe we should wrap this into an error?

hey, thanks for your reply! ok I see what you mean but the function to calculate the lambda is not public so is not necessarily useful to use as a library. I think this should be done in another PR, right?

xJonathanLEI:
I haven't benchmarked but my gut sense says this refactor would hurt performance.

xJonathanLEI:
Thanks for the PR but I'm not convinced this is good refactor. You're taking 2 code paths, putting them into a function, which itself is just 2 completely separated paths. It does not help with code readability IMO and suffers from the issue that @tcoratger brought up (thanks for the review btw). I'm closing this PR now.

PR #549, opened 2024-02-17
Add starkent transaction retrieval example

DoTheBestToGetTheBest:
This PR introduces a new example into our `examples/` crates, focusing on demonstrating how to retrieve transaction details.


DoTheBestToGetTheBest:
> Hi sorry but I think as an example the encapsulation being added here isn't very helpful. Also IMO probably not the best idea to do error handling, as it makes the example more complicated than it needs to be, and loses the focuse of what matters.

Are you okay if i open another pr with no encapsulation ?

PR #550, opened 2024-02-17
Improve `FieldElement` struct construction

tcoratger:
## Description
This pull request refactors the `FieldElement` structure to remove the unnecessary `inner` field, enhancing efficiency and minimizing struct size. Additionally, it implements the `std::ops::Deref` and `std::ops::DerefMut` traits to facilitate standard struct manipulations.

## Changes
- Removed the `inner` field from the `FieldElement` structure.
- Implemented `std::ops::Deref` and `std::ops::DerefMut` traits for improved struct operations.

## Impact
- Improved efficiency by eliminating redundant struct fields.
- Reduced struct size for better memory optimization.
- Simplified struct manipulation with the addition of trait implementations.

### DIFF
--- a/starknet-ff/src/lib.rs
+++ b/starknet-ff/src/lib.rs
@@ -178,7 +178,7 @@ impl FieldElement {
         }
 
         Fr::from_bigint(u256_to_biginteger256(&res))
-            .map(|inner| Self(inner))
+            .map(Self)
             .ok_or(FromStrError::OutOfRange)
     }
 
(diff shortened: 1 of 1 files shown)


tcoratger:
> TBH I don't think the `inner` part makes any difference. I could be wrong but I think a single field struct should be exactly equivalent to a newtype struct in terms of performance. So changing that for better performance doesn't feel justified.

This one doesn't seem really relevant since we have this one in the pipeline: https://github.com/xJonathanLEI/starknet-rs/pull/562 which will delete all of this.

xJonathanLEI:
Yeah that's true.

xJonathanLEI:
`starknet-ff` was removed as part of #562. Closing this now as it's no longer relevant.

PR #552, opened 2024-02-18
fix `add_assign` impl between projective points

tcoratger:
During `add_assign` implementation, z1 and z2 don't have to be equal. As a result, when `u0 == u1`, simply testing ` self.y == rhs.y` to confirm that a point doubling is needed is false. 

If we have y1 = y2, z1 != z2 can lead to `self.y * rhs.z != rhs.y * self.z`, hence identity needs to be returned. 

This PR fixes this.


tcoratger:
Can be closed with https://github.com/xJonathanLEI/starknet-rs/pull/562

PR #553, opened 2024-02-19
Improve projective mul by scalar perf

tcoratger:
During the multiplication of a projective point by a scalar, the leading zeros can be skipped in the loop, because they induce unnecessary doubling calculations, giving always the identity value back before the first non zero element is reached in the `rhs` boolean array.

Skipping this useless calculations can reduce drastically the calculation costs, especially for small numbers (for example, is our scalar is 1, then 255 useless doubling are performed).


xJonathanLEI:
Rebasing on `master` for CI.

PR #558, opened 2024-02-28
feat: sequencer error code `VALIDATE_FAILURE`

jbcaron:
This PR add the error code `StarknetErrorCode.VALIDATE_FAILURE` from sequencer API.


xJonathanLEI:
Thanks for the PR. However, `starknet-core/src/types/codegen.rs` is for RPC types, which is strictly derived from the specification and must not be manually edited.

The RPC error type already contains a `ValidationFailure` variant, which you can use when converting from sequencer error to RPC error.

PR #559, opened 2024-02-28
fix(class): 🐛 `LegacyContractClass` deserialization failing on empty ABI

Trantorian1:
# Pull Request type

- Bugfix

## What is the current behavior?

[`LegacyContractClass`](https://github.com/xJonathanLEI/starknet-rs/blob/400deb6f058447e8b67316a92ecd9ad3b6d4330c/starknet-core/src/types/contract/legacy.rs#L31) deserialization fails on classes with a missing ABI. 

Earliest example of this is class [`0x01cbbbd9543d07a5769e0d17bca0ede054082a27689aa7b6b3731851e0c65326`](https://starkscan.co/class/0x0371b5f7c5517d84205365a87f02dcef230efa7b4dd91a9e4ba7e04c5b69d69b#overview) at block [`20732`](https://starkscan.co/block/20732).

## What is the new behavior?

- `LegacyContractClass` can now have a missing (`Option`) ABI.

## Does this introduce a breaking change?

Yes. Code that retrieves the ABI from a `LegacyContractClass` will have to be updated to handle the new `Option`.


xJonathanLEI:
How are you parsing the class? IIRC the class as generated by the compiler will always have the `abi` field, even when it's empty. This type is for the compiler output. If you're parsing response from providers, you should use `starknet_core::types::ContractClass`.

PR #563, opened 2024-03-08
fix nightly imports

tcoratger:
Related: https://doc.rust-lang.org/std/prelude/

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,35 +1,41 @@
 on:
   push:
     branches:
-      - master
+      - "master"
   pull_request:
 
-name: Linting
+name: "Linting"
 
 jobs:
-  lint:
-    name: Lints
-    runs-on: ubuntu-latest
+  format:
+    name: "Check code format"
+    runs-on: "ubuntu-latest"
     steps:
-      - name: Checkout source code
-        uses: actions/checkout@v2
-      - name: Use Rust cache
-        uses: Swatinem/rust-cache@v1
+      - name: "Checkout source code"
+        uses: "actions/checkout@v3"
+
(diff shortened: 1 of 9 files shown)


xJonathanLEI:
Hi I don't quite understand this PR. When does it mean to "fix nightly imports"? The library should already work with nightly? It's checked on CI that nightly builds are successful. Or is it not working for `no_std` on nightly now?

tcoratger:
> Hi I don't quite understand this PR. When does it mean to "fix nightly imports"? The library should already work with nightly? It's checked on CI that nightly builds are successful. Or is it not working for `no_std` on nightly now?

@xJonathanLEI This is related to nightly target. In the CI and in the repo we are targeting Rust nigthly. Recently, new components have been added to the Rust `std` prelude, especially for nightly (https://doc.rust-lang.org/std/prelude/index.html) such as `Vec` for example. So there is a compilation warning on the reimports of these types now integrated into the prelude (you can see it on the CIs of the other PRs in progress). As we are targeting `no_std` on certain builds, we cannot simply remove these imports because they will no longer be recognized at all during `no_std` builds. Consequently, the simplest thing is to hide the import of the `Vec` in a `vec::*`, which is done in this PR for example. And as a result of that the full CI is ok.

xJonathanLEI:
I see. So the problem is that we only run tests on nightly for CI, but not clippy. Clippy is currently only run for stable Rust. If we want to fix what's proposed in this PR we better also have it on CI to enforce that too tho.

xJonathanLEI:
Again, this PR changes `codegen.rs`, which is not supposed to happen. That file must not be manually edited.

xJonathanLEI (approved):

PR #566, opened 2024-03-10
Replace `ELEMENT_UPPER_BOUND` by `FieldElement::MAX` in ecdsa

tcoratger:
**Description:**
In the `FieldElement` module, we've already defined a `Max` field to indicate the maximum value a field element can take before the modulus is reached, specifically equal to 2^251 + 17 * 2^192. Consequently, there's no need to redefine this value in the crypto ecdsa file. By leveraging the existing definition from `FieldElement`, we can ensure consistency and avoid redundancy in our codebase.


xJonathanLEI:
Yep yep. IIRC this was done specifically to align with the official impl from the StarkWare C++ lib back in the day.

PR #569, opened 2024-03-14
Improve poseidon hash perf by using assign operators

tcoratger:
This pull request aims to enhance the performance of the poseidon hash by utilizing assignment multiplication (`mul_assign`) and addition (`add_assign`) operators whenever possible. By doing so, unnecessary copies are avoided, resulting in improved overall performance.


xJonathanLEI:
Rebasing onto `master` for CI.

xJonathanLEI:
Running the benchmark on my machine the perf changes from this seems to be within noise threshold. I'm not sure about this. Need to test a bit more I guess.

xJonathanLEI:
@tcoratger This PR seems no longer relevant as the new `Felt` type does not provide methods equivalent to what you're adding here?

xJonathanLEI:
Also it will be superseded by anything that comes out of #608 anyways. Closing now.

PR #580, opened 2024-04-19
Add `BraavosAccountFactory`

ddoktorski:
The code has been taken from the [starkli implementation](https://github.com/xJonathanLEI/starkli/blob/53ffb435fff917a245e3ef5e417a73299a742730/src/account_factory/braavos.rs). It seems reasonable to include all account factories within `starknet-rs`. Once it is added here, it can be removed from `starkli`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1660,6 +1660,7 @@ dependencies = [
  "serde",
  "serde_json",
  "starknet-core",
+ "starknet-crypto",
  "starknet-providers",
  "starknet-signers",
  "thiserror",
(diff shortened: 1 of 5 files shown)


xJonathanLEI:
Thanks for the PR. The Braavos account contract factory implementation from Starkli, which you're bringing over with this PR, is rather incomplete as it doesn't support deploying with multiple signers, which is, if I understand correctly, _the_ core idea of a Braavos account. I don't think in its current form I'm comfortable with committing to an API.

Therefore I'm closing this. If you really need the factory you can always implement it in your own code until `starknet-rs` will eventually provide a canonical implementation.

PR #590, opened 2024-05-21
feat: U256 type

xJonathanLEI:
The `u256` Cairo type equivalent for `starknet-rs`.


xJonathanLEI:
This is quite preliminary and missing many important trait impls. Merging this first to have something we can build on.

PR #594, opened 2024-06-03
feat: make hasher accept iterators rather than arrays

### LINKED ISSUE #575: `compute_hash_on_elements` could accept any iterator on `FieldElement`
https://github.com/xJonathanLEI/starknet-rs/blob/d980869d44ef86249b4077f4ec809165f470678d/starknet-core/src/crypto.rs#L49

Right now the signature requires the input to be `&[FieldElement]` but it's implementation would still be valid if the input were to be a more generic `T: Iter<FieldElement>`. It would allow more flexibility for the lib users

tdelabro:
closes #575

No breaking change as both new impl are backward compatible with the only previously allowed input type `&[FieldElement]`

### DIFF
--- a/starknet-core/src/crypto.rs
+++ b/starknet-core/src/crypto.rs
@@ -46,14 +46,19 @@ mod errors {
 }
 pub use errors::{EcdsaSignError, EcdsaVerifyError};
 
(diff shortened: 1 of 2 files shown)


tcoratger (approved):

tcoratger:
Lgtm pending @xJonathanLEI

Issue #593, opened 2024-05-29
Extend JSON-RPC transport capabilities for batch requests


tcoratger:
#### Context

Currently, the `JsonRpcRequest` structure in

https://github.com/xJonathanLEI/starknet-rs/blob/master/starknet-providers/src/jsonrpc/transports/http.rs

allows for making JSON-RPC requests with a specified method. To support the needs of various RPC services, it would be beneficial to extend the transport capabilities to handle batch requests. This can likely be achieved by improving the `JsonRpcTransport` trait.

#### Proposed Changes

1. **Enhance `JsonRpcTransport` Trait**:
   - Modify the [`JsonRpcTransport` trait](https://github.com/xJonathanLEI/starknet-rs/blob/945145967187c8fd1c2679aebd23f1f5fc83f9b6/starknet-providers/src/jsonrpc/transports/mod.rs#L14-L25) to include a new `send_requests` method that facilitates sending requests in batch.

2. **Introduce `JsonRpcRequests` Type**:
   - Create a new type, `JsonRpcRequests`, which is essentially a vector of `JsonRpcRequest`.

The previous two points are suggestions only. Feel free to propose any other way to achieve the same goal in a better way.

#### Additional Resources

For more details on batch requests, refer to the [associated documentation](https://sajya.github.io/docs/batch/).

thetheveloper:
## Current approach
The @tcoratger idea is a correct approach, however we believe it could be simplified a bit more. Unless there is a specific reason/use case for introducing a new type like `JsonRpcRequests`,  we could just define a vector of `JsonRpcRequest`. The rest would be the same as suggested in the original Issue description. 

## Implementation details 
For now the function `send_requests` can be implemented in the `JsonRpcTransport` struct and called with vector of requests as an argument. It’s easy to pass requests of the same `JsonRpcMethod`. 

## Improvement proposal
An improvement to this implementation is up for discussion and a separate issue. In essense, if we want to make a batch request of different RPC methods we need to implement enum of `JsonRpcRequest` params to allow vector to store different data types. 

The general idea would be the following:
We introduce an `enum` type that allows all possible payloads to be passed through the generic type.
_The only limitation here would be that whenever there is a new request type, the enum needs to be expanded in the library, which is not very robust - hence just this is a topic for separate discussion.

## Summary
If the presented approach in the first 2 paragraphs is compatible with the lib, **we already have the code working and we would like to share it in a new PR** to be verified by the @xJonathanLEI

PR #600, opened 2024-06-15
feat: implement JsonRpc transport for batch requests

thetheveloper:
Introduces `JsonRpc` batching requests described in #593

### DIFF
--- a/examples/jsonrpc_batch.rs
+++ b/examples/jsonrpc_batch.rs
@@ -1,0 +1,57 @@
+use starknet::{
+    core::types::{BlockId, BlockTag},
+    providers::{
+        jsonrpc::{HttpTransport, JsonRpcClient},
+        Provider, Url,
+    },
+};
+use starknet_core::types::requests::{GetBlockWithTxHashesRequestRef, GetBlockWithTxsRequestRef};
+use starknet_providers::jsonrpc::{JsonRpcMethod, JsonRpcRequestParams};
+
+#[tokio::main]
+async fn main() {
+    // Create a new JSON RPC client using HTTP transport with the specified URL
+    let provider = JsonRpcClient::new(HttpTransport::new(
+        Url::parse("https://starknet-sepolia.public.blastapi.io/rpc/v0_7").unwrap(),
+    ));
+
+    // batch_requests allows to define a vector of requests for batch processing, ensuring each request specifies its corresponding JsonRpcMethod and JsonRpcRequestParams.
+    // This approach allows for a generic way to handle batch requests.
+    let batch_mixed_results = provider
+        .batch_requests(vec![
+            // Request 1: Retrieve block data including transaction hashes.
+            (
+                JsonRpcMethod::GetBlockWithTxHashes,
+                JsonRpcRequestParams::GetBlockWithTxHashes(GetBlockWithTxHashesRequestRef {
+                    block_id: BlockId::Tag(BlockTag::Latest).as_ref(),
+                }),
+            ),
+            // Request 2: Retrieve block data including full transaction details.
+            (
+                JsonRpcMethod::GetBlockWithTxs,
+                JsonRpcRequestParams::GetBlockWithTxs(GetBlockWithTxsRequestRef {
+                    block_id: BlockId::Tag(BlockTag::Latest).as_ref(),
+                }),
+            ),
+        ])
+        .await;
+
+    match batch_mixed_results {
+        Ok(v) => println!("{v:#?}"),
+        Err(e) => println!("Error: {e:#?}"),
+    }
+
+    // The following example demonstrates the process of sending a batch request to retrieve multiple blocks, each including transaction hashes.
+    // get_block_with_tx_hashes_batch utilizes a vector of BlockId parameters to construct the batch request.
+    let batched_blocks = provider
+        .get_block_with_tx_hashes_batch(vec![
+            BlockId::Tag(BlockTag::Latest),
+            BlockId::Tag(BlockTag::Latest),
+        ])
+        .await;
+
+    match batched_blocks {
+        Ok(v) => println!("{v:#?}"),
+        Err(e) => println!("Error: {e:#?}"),
+    }
+}
--- a/starknet-providers/src/any.rs
+++ b/starknet-providers/src/any.rs
@@ -1,4 +1,5 @@
 use async_trait::async_trait;
+use serde::Serialize;
 use starknet_core::types::{
     BlockHashAndNumber, BlockId, BroadcastedDeclareTransaction,
     BroadcastedDeployAccountTransaction, BroadcastedInvokeTransaction, BroadcastedTransaction,
@@ -11,7 +12,7 @@ use starknet_core::types::{
 };
 
 use crate::{
-    jsonrpc::{HttpTransport, JsonRpcClient},
+    jsonrpc::{HttpTransport, JsonRpcClient, JsonRpcMethod},
     Provider, ProviderError, SequencerGatewayProvider,
 };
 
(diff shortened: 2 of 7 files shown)


starknet-providers/src/any.rs
```diff
@@ -667,4 +668,45 @@ impl Provider for AnyProvider {
...
+        match self {
+            Self::JsonRpcHttp(inner) => {
+                <JsonRpcClient<HttpTransport> as Provider>::batch_requests(inner, requests).await
+            }
+            Self::SequencerGateway(inner) => {
+                <SequencerGatewayProvider as Provider>::batch_requests(inner, requests).await
+            }
+        }
+    }
+    async fn get_block_with_tx_hashes_batch<B>(
```
tcoratger, line 689: fmt nit: can we have a space between the two functions?

starknet-providers/src/any.rs
```diff
@@ -667,4 +668,45 @@ impl Provider for AnyProvider {
...
+        &self,
+        requests: I,
+    ) -> Result<Vec<serde_json::Value>, ProviderError>
+    where
+        I: IntoIterator<Item = (JsonRpcMethod, P)> + Send + Sync,
+        P: Serialize + Send + Sync,
+    {
+        match self {
+            Self::JsonRpcHttp(inner) => {
+                <JsonRpcClient<HttpTransport> as Provider>::batch_requests(inner, requests).await
```
tcoratger, line 682:
```suggestion
                inner.batch_requests(requests).await
```

starknet-providers/src/any.rs
```diff
@@ -667,4 +668,45 @@ impl Provider for AnyProvider {
...
+            }
+            Self::SequencerGateway(inner) => {
+                <SequencerGatewayProvider as Provider>::batch_requests(inner, requests).await
```
tcoratger, line 685:
```suggestion
                inner.batch_requests(requests).await
```

starknet-providers/src/any.rs
```diff
@@ -667,4 +668,45 @@ impl Provider for AnyProvider {
...
+    ) -> Result<Vec<MaybePendingBlockWithTxHashes>, ProviderError>
+    where
+        B: AsRef<BlockId> + Send + Sync,
+    {
+        match self {
+            Self::JsonRpcHttp(inner) => {
+                <JsonRpcClient<HttpTransport> as Provider>::get_block_with_tx_hashes_batch(
+                    inner, block_ids,
+                )
+                .await
```
tcoratger, line 701:
```suggestion
                inner.get_block_with_tx_hashes_batch(
                   block_ids,
                )
                .await
```

starknet-providers/src/any.rs
```diff
@@ -667,4 +668,45 @@ impl Provider for AnyProvider {
...
+            }
+            Self::SequencerGateway(inner) => {
+                <SequencerGatewayProvider as Provider>::get_block_with_tx_hashes_batch(
+                    inner, block_ids,
+                )
+                .await
```
tcoratger, line 707:
```suggestion
               inner.get_block_with_tx_hashes_batch(
                   block_ids,
                )
                .await
```

starknet-providers/src/jsonrpc/mod.rs
```diff
@@ -204,6 +216,31 @@ where
             }
         }
     }
+    async fn send_requests<I, P, R>(&self, requests: I) -> Result<Vec<R>, ProviderError>
```
tcoratger, line 219: fmt nit: can we have a space between the two functions?

starknet-providers/src/jsonrpc/mod.rs
```diff
@@ -91,13 +91,25 @@ pub enum JsonRpcMethod {
...
 #[derive(Debug, Clone)]
 pub struct JsonRpcRequest {
     pub id: u64,
     pub data: JsonRpcRequestData,
 }
 
 #[derive(Debug, Clone)]
+pub struct JsonRpcRequests {
+    pub requests: Vec<JsonRpcRequest>,
+}
```
tcoratger, line 110:
```suggestion
pub struct JsonRpcRequests (
   pub Vec<JsonRpcRequest>
);
```

starknet-providers/src/jsonrpc/transports/http.rs
```diff
@@ -101,4 +101,46 @@ impl JsonRpcTransport for HttpTransport {
...
+
+        let serialized_batch = serde_json::to_string(&batch_requests).map_err(Self::Error::Json)?;
+
+        let mut request = self
+            .client
+            .post(self.url.clone())
+            .body(serialized_batch)
+            .header("Content-Type", "application/json");
+
+        for (name, value) in self.headers.iter() {
```
tcoratger, line 133:
```suggestion
        for (name, value) in &self.headers {
```

examples/jsonrpc_batch.rs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    let batched_blocks = provider.get_block_with_tx_hashes_batch(vec![BlockId::Tag(BlockTag::Latest), BlockId::Tag(BlockTag::Latest)]).await;
+
+    match batched_blocks {
+        Ok(v) => println!("{v:#?}"),
+        Err(e) => println!("Error: {e:#?}"),
+    }
+
+}
```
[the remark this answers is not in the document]
thetheveloper: Good point. Added those

starknet-providers/src/jsonrpc/mod.rs
```diff
@@ -218,6 +255,35 @@ where
...
+        B: AsRef<BlockId> + Send + Sync,
+    {
+        let requests = block_ids.iter().map(|b_id| {
+            (
+                JsonRpcMethod::GetBlockWithTxHashes,
+                GetBlockWithTxHashesRequestRef {
+                    block_id: b_id.as_ref(),
+                },
+            )
+        });
```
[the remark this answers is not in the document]
thetheveloper: Good catch, agree full name makes much more sense.

starknet-providers/src/jsonrpc/transports/http.rs
thetheveloper: Fixed.

starknet-providers/src/jsonrpc/mod.rs
thetheveloper: Fixed.

starknet-providers/src/any.rs
thetheveloper: Fixed.

starknet-providers/src/any.rs
thetheveloper: Fixed.

starknet-providers/src/any.rs
thetheveloper: Fixed.

starknet-providers/src/jsonrpc/transports/http.rs
```diff
@@ -101,4 +101,46 @@ impl JsonRpcTransport for HttpTransport {
...
+        }
+
+        let response = request.send().await.map_err(Self::Error::Reqwest)?;
+
+        let response_body = response.text().await.map_err(Self::Error::Reqwest)?;
+        trace!("Response from JSON-RPC: {}", response_body);
+
+        let parsed_response = serde_json::from_str(&response_body).map_err(Self::Error::Json)?;
+
+        Ok(parsed_response)
```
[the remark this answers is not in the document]
thetheveloper: Fixed.

thetheveloper:
Thank you @tcoratger for the review. All review comments addresses in the latest commit.

starknet-providers/src/provider.rs
```diff
@@ -298,6 +301,21 @@ pub trait Provider {
             Err(ProviderError::ArrayLengthMismatch)
         }
     }
+
+    async fn batch_requests<I, P>(
+        &self,
+        requests: I,
+    ) -> Result<Vec<serde_json::Value>, ProviderError>
```
xJonathanLEI, line 308: Do we have to resort to `serde_json::Value` here? I generally dislike having that as part of the interface.

thetheveloper: Good question. Not necessarily, however this function was specifically introduced for this approach - to return generic JSON type based on the type returned. Otherwise, we'd have to create custom types for this and that would limit the ways we can call it

It's possible to change it if you prefer any other direction

tcoratger:
@thetheveloper can you rebase to trigger the ci again? I think this should be good now

thetheveloper:
> @thetheveloper can you rebase to trigger the ci again? I think this should be good now

@tcoratger rebase done. I tried to include the test fixes (hence last commit), but from what if the tests are not run one after another try everything is fine. I suppose RPC node needs to process transactions from the previous iteration of testing, so the next one doesn't fail

thetheveloper:
Should be fine now. Latest base is in place @tcoratger

adrienlacombe:
@tcoratger @xJonathanLEI anything else needed? thanks!

tcoratger:
> @tcoratger @xJonathanLEI anything else needed? thanks!

For me this is ok to merge but as the PR is quite big and important I would appreciate another review from @xJonathanLEI if possible?

xJonathanLEI:
Squashed and rebased.

adrienlacombe:
@tcoratger maybe you can review as well? ty ty

starknet-providers/src/provider.rs
adrienlacombe: ty @tcoratger , @xJonathanLEI can you please let us know your thoughts? ty

xJonathanLEI:
While this implementation probably works, I'm really not a fan of adding `serde_json::Value` to a supposedly neutral trait of `Provider`. I mean it's probably acceptable to add it to `JsonRpcTransport`, because as its name suggests, it deals with JSON.

I know this is a bit not very pragmatic - after all, `Provider` _is_ modelled after the JSON-RPC standard. However, with the current design any `Provider` user would now have to deal with deserialization... which is far from ideal.

Anyway, I just prefer a design where we don't have to pollute `Provider` with JSON stuff.

The diffs are also a bit all over the place:

- Why do we have a `get_block_with_tx_hashes_batch` method? Wouldn't that be covered by `send_requests`?
- Same goes for `JsonRpcRequestParams` - this type only covers 2 methods.

Closing this as I'll work on an alternative version. Thanks for working on this.

PR #605, opened 2024-06-18
feat: ledger signer support

xJonathanLEI:
The very first attempt at adding Ledger support. The Ledger support is feature-gated by a `ledger` feature on both the `starknet-signers` and `starknet` crates. It's turned off by default.

The current Ledger app for Starknet is quite limited: it only supports blind signing. Nevertheless, it's a huge step forward compared to software keystores.

This PR adds 2 examples that work as long as you have Ledger plugged in, unlocked, and the Starknet app open. Test with:

```console
cargo run --example ledger_public_key
cargo run --example deploy_account_with_ledger
```

Major caveats of the current implementation:

1. Confirming the public key on the device's screen is not supported;
2. Higher-level types like `Account` and `AccountFactory` still hold the assumption that all signers are non-interactive, and would request signatures even just for fee estimation; this results in excessive signing prompts on the Ledger device;
3. The EIP-2645 path is difficult to write correctly without copy pasting; maybe we should add a type that encapsulates this and converts into `DerivationPath`.

That said, if anyone wants to start using it from this PR, they totally can and it's 100% functional :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.22.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
 name = "adler"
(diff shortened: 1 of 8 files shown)


xJonathanLEI:
Pathfinder bug is causing test failures: https://github.com/eqlabs/pathfinder/issues/2085

PR #612, opened 2024-06-20
Enable `num-traits` feature in `starknet-types-core`

kariy:
Before the migration to the new felt type, we used to be able to do `felt.try_into<u128>()` but the implementation has been removed since. The `num-traits` feature pretty much add implementations for the conversion to primitive uint types.

Would be nice to have it enabled directly from `starknet-rs` instead of having to add the `starknet-types-core` crate manually.


xJonathanLEI (approved):
LGTM

Issue #609, opened 2024-06-18
replace Pedersen implementation by `type-rs`


tcoratger:
The following Pedersen implementation

https://github.com/xJonathanLEI/starknet-rs/blob/4dbc1692d4f295de803484181e07803201528922/starknet-crypto/src/pedersen_hash.rs

should be replaced by the dedicated `type-rs` provided implementation:

https://github.com/starknet-io/types-rs/blob/main/crates/starknet-types-core/src/hash/pedersen.rs

thetheveloper:
We would like to be assigned to this issue

cc @tcoratger @xJonathanLEI

tcoratger:
> We would like to be assigned to this issue
> 
> cc @tcoratger @xJonathanLEI

Assigned

PR #614, opened 2024-06-23
feat: replace Pedersen implementation with type-rs

thetheveloper:
Closes https://github.com/xJonathanLEI/starknet-rs/issues/609 

Removes current Pedersen implementation & replaces with the one implemented in `type-rs`

cc @tcoratger @xJonathanLEI please review and check if this is what the #609 meant (that's our understanding).


tcoratger:
@xJonathanLEI I see these kind of errors in the CI in different prs, is it normal? I think we have something strange here, not linked to the pr, wdyt?

xJonathanLEI:
Hi @tcoratger there was a node failure of the Sepolia node I hosted for running the tests. It should be all good now.

tcoratger (approved):

PR #625, opened 2024-07-05
bump `serde_with` to 3.8

tcoratger:


xJonathanLEI:
Note that this is a breaking change as `serde_with` ends up being part of the public API. But this is fine and the next release will be breaking anyway.

xJonathanLEI (approved):

xJonathanLEI:
Hmm all I did was nudging the version further to `3.9` as that's released today. Then it fails. Looks like `serde_with` somehow breaks semver?

PR #630, opened 2024-07-16
feat: replace poseidon impl with type-rs

### LINKED ISSUE #608: replace Poseidon implementation by `types-rs`
The following Poseidon implementation

https://github.com/xJonathanLEI/starknet-rs/blob/4dbc1692d4f295de803484181e07803201528922/starknet-crypto/src/poseidon_hash.rs

should be replaced by the dedicated `type-rs` provided implementation:

https://github.com/starknet-io/types-rs/blob/main/crates/starknet-types-core/src/hash/poseidon.rs

thetheveloper:
Closes #608 

Removes current Poseidon implementation & replaces with the one implemented in type-rs

### DIFF
--- a/starknet-crypto/src/lib.rs
+++ b/starknet-crypto/src/lib.rs
@@ -20,9 +20,7 @@ pub use starknet_types_core::felt::Felt;
 
 pub use pedersen_hash::pedersen_hash;
 
-pub use poseidon_hash::{
-    poseidon_hash, poseidon_hash_many, poseidon_hash_single, poseidon_permute_comp, PoseidonHasher,
-};
+pub use poseidon_hash::{poseidon_hash, poseidon_hash_many, poseidon_hash_single, PoseidonHasher};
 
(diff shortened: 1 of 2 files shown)


thetheveloper:
@tcoratger the PR is ready for review 

`ci` failed for the same reason the other PRs failed - locally everything was fine, pushing here again it started failing mostly with `InvalidNonce` or `Provider(StarknetError(DuplicateTx))`. Could you please take a look if the implementation is in fact correct as all tests are passing locally?

xJonathanLEI:
Somehow these two tests would always say duplicate tx:

- `can_execute_eth_transfer_invoke_v1_with_jsonrpc`
- `can_execute_eth_transfer_invoke_v3_with_jsonrpc`

It's almost like the sequencer refuses to process those txs and they got stuck.

xJonathanLEI:
I manually bumped the gas price to use that nonce. Should be good again.

tcoratger (approved):

PR #653, opened 2024-09-10
feat: jsonrpc batch requests

### LINKED ISSUE #593: Extend JSON-RPC transport capabilities for batch requests

xJonathanLEI:
Resolves #593. Supersedes #600.

Implements a type-safe version of the JSON-RPC batch request feature. Users can now do this:

```rust
let responses = provider
    .batch_requests([
        ProviderRequestData::BlockNumber(BlockNumberRequest),
        ProviderRequestData::GetBlockTransactionCount(GetBlockTransactionCountRequest {
            block_id: BlockId::Number(100),
        }),
    ])
    .await
    .unwrap();

match (&responses[0], &responses[1]) {
    (
        ProviderResponseData::BlockNumber(block_number),
        ProviderResponseData::GetBlockTransactionCount(count),
    ) => {
        println!("The latest block is #{}", block_number);
        println!("Block #100 has {} transactions", count);
    }
    _ => panic!("unexpected response type"),
}
```

and the underlying requests would be sent in the just one HTTP request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,19 +92,21 @@ Examples can be found in the [examples folder](./examples):
 
 6. [Query the latest block number with JSON-RPC](./examples/jsonrpc.rs)
 
-7. [Call a contract view function](./examples/erc20_balance.rs)
+7. [Batched JSON-RPC requests](./examples/batch.rs)
 
-8. [Deploy an Argent X account to a pre-funded address](./examples/deploy_argent_account.rs)
+8. [Call a contract view function](./examples/erc20_balance.rs)
 
-9. [Inspect public key with Ledger](./examples/ledger_public_key.rs)
+9. [Deploy an Argent X account to a pre-funded address](./examples/deploy_argent_account.rs)
 
-10. [Deploy an OpenZeppelin account with Ledger](./examples/deploy_account_with_ledger.rs)
+10. [Inspect public key with Ledger](./examples/ledger_public_key.rs)
 
-11. [Transfer ERC20 tokens with Ledger](./examples/transfer_with_ledger.rs)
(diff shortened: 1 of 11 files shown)


xJonathanLEI:
Note: technically we can also implement a method that accepts only up to 32 requests but is truly type-safe by leveraging generic params. Something similar to:

```rust
let (block_number, count) =
  provider.batch_requests(BlockNumberRequest, GetBlockTransactionCountRequest {...}).await.unwrap();
```

which would be much nicer for use cases where the request types are statically known at compile time.

Let's settle for this for now though.

PR #664, opened 2024-10-03
bump dependencies and lock file

tcoratger:
This PR:
- Bumps some deps
- Let the system find itself the appropriate versioning instead of specifying an exact version


tcoratger:
cc @xJonathanLEI

xJonathanLEI:
As communicated separately, there seems to be no good reason to tweak version ranges like this.

PR #665, opened 2024-10-04
bump starknet-types-core to 0.1.7

tcoratger:


xJonathanLEI:
As communicated separately, I believe it's a good practice to specify the minimum version of dependency we need instead of always chasing the latest version to be min, hence unnecessarily forcing downstream to use newer versions (sometimes newer versions might break people for whatever reason).

PR #700, opened 2025-03-05
Fix runtime error during transaction hash calculation

marioiordanov:
in `PreparedAccountDeploymentV3::transaction_hash()` resource buffer for l1_data_gas property is incorrectly filled 2 times with l1_data_gas


xJonathanLEI (approved):
Nice catch. This part is not tested as it requires a fresh address to be funded, so it's hard to set up properly.

PR #704, opened 2025-03-20
feat: port starknet.js SNIP-12 deviations

xJonathanLEI:
`starknet-rs` currently ships a rather correct implementation of [SNIP-12](https://github.com/starknet-io/SNIPs/blob/be4423ab0a6bf0678e8d5a66c712dd160b47334c/SNIPS/snip-12.md). Unfortunately, the de-facto standard `starknet.js` ships multiple bugs: https://github.com/starknet-io/starknet.js/pull/1292.

Despite all 3 fixes being merged, only the enum dependency fix (https://github.com/starknet-io/starknet.js/pull/1289) is currently expected to ever be released. Therefore, this "alignment" PR uses message hashes generated from `starknet.js` [v6.24.0](https://github.com/starknet-io/starknet.js/releases/tag/v6.24.0) _plus a patch of https://github.com/starknet-io/starknet.js/pull/1289_.

> [!NOTE]
>
> The test case `test_message_hash_v1_with_simple_enum` is indifferent to whether the patch is applied or not. Only the `test_message_hash_v1_with_enum_nested` case is affected.

After this PR, the two libraries should generate the same message hashes in most cases involving enums. For the rather uncommon cases where dependency types are used in enum data, we would need starknet.js to release https://github.com/starknet-io/starknet.js/pull/1289.

### DIFF
--- a/starknet-core/src/types/typed_data/mod.rs
+++ b/starknet-core/src/types/typed_data/mod.rs
@@ -878,7 +878,52 @@ mod tests {
 
     #[test]
     #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
-    fn test_message_hash_v1_with_enum() {
+    fn test_message_hash_v1_with_simple_enum() {
+        let raw = r###"{
+  "types": {
+    "StarknetDomain": [
+      { "name": "name", "type": "shortstring" },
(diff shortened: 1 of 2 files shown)


xJonathanLEI:
cc @penovicp

PR #713, opened 2025-04-02
feat(core/types/Call): derive serde on struct Call

### LINKED ISSUE #712: derive serde on Call
https://github.com/xJonathanLEI/starknet-rs/blob/master/starknet-core/src/types/call.rs

This type is not serializable. This prevents me from easily preparing transaction content in some tool/service before passing it to the wallet to sign and send.
I'm thinking about usecase like deep-link

tdelabro:
closes #712


tdelabro:
While at it I also added the low-hanging fruits: `PartialEq, Eq, Hash`

Imo, this kind of derive impl should be all over the codebase, on every type you define where it is possible to derive.
This is a lib, you don't know what users will need, so you should cover as many things as possible.

Maybe behind a feature flag for `serde`

xJonathanLEI:
Hey thanks for the PR but could you remove `.DS_STORE`?

tdelabro:
@xJonathanLEI yup sorry about that. I added those to `.gitignore` too

tdelabro:
@xJonathanLEI ready to merge

xJonathanLEI (approved):

PR #735, opened 2025-05-28
Use `zeroize` crate for SigningKey.

FilipLaurentiu:
`fn secret_scalar` returns a reference to the secret scalar so the value is still zeroed on drop

Update `starknet-types-core` version to `0.1.8` everywhere to use the `zeroize` feature https://github.com/starknet-io/types-rs/commit/67468ff2bb725e1601435ea340e5b0209e6c1876


starknet-crypto/Cargo.toml
```diff
@@ -24,7 +24,7 @@ rfc6979 = { version = "0.4.0", default-features = false }
 sha2 = { version = "0.10.6", default-features = false }
 zeroize = { version = "1.6.0", default-features = false }
 hex = { version = "0.4.3", default-features = false, optional = true }
-starknet-types-core = { version = "0.1.7", default-features = false, features = ["curve", "hash"] }
+starknet-types-core = { version = "0.1.8", default-features = false, features = ["curve", "hash"] }
```
FilipLaurentiu, line 27: Latest version of `starknet-types-core` support operations on EC points as well https://github.com/starknet-io/types-rs/commit/e0eff28053b0280629c3bf6827e6246995a27503

starknet-signers/src/key_pair.rs
```diff
@@ -100,8 +101,8 @@ impl SigningKey {
     }
 
     /// Gets the secret scalar in the signing key.
-    pub const fn secret_scalar(&self) -> Felt {
-        self.secret_scalar
+    pub const fn secret_scalar(&self) -> &Felt {
```
glihm, line 104: What motivated a change to `&` when `Felt` is copiable? Most usage takes a `Felt` as argument which would lead to a copy while dereferencing.

FilipLaurentiu: For this case, the motivation is to avoid creating another copy of a Felt that will not be zeroized on drop and stay in memory. Returning just a reference to the internal value will solve this problem, the value will still be zeroized at the end.

FilipLaurentiu: > What motivated a change to `&` when `Felt` is copiable? Most usage takes a `Felt` as argument which would lead to a copy while dereferencing.

For this case, the motivation is to avoid creating another copy of a Felt that will not be zeroized on drop and stay in memory. Returning just a reference to the internal value will solve this problem, the value will still be zeroized at the end.

PR #747, opened 2025-06-26
:bug: Fix typed data hash

florian-bellotti:


starknet-core/src/types/typed_data/encoder.rs
```diff
@@ -342,8 +342,12 @@ impl Encoder {
                     }
                 };
 
-                get_selector_from_name(str_value)
-                    .map_err(|_| TypedDataError::InvalidSelector(str_value.to_owned()))?
+                match str_value.strip_prefix("0x") {
```
xJonathanLEI, line 345: It appears that parsing a selector regardless of prefix (existing behavior) is the expected behavior given that the type is literally `Selector`.

IIUC this change here is to make `starknet-rs` deviate from SNIP-12 to be compatible with the de-facto standard `starknet.js`? If so could you please annoate the same way as other such deviations in the codebase?

starknet-core/src/types/typed_data/mod.rs
```diff
@@ -491,6 +491,59 @@ mod tests {
         );
     }
 
+    #[cfg(test)]
+    mod message_hash {
```
xJonathanLEI, line 495: There seems to be no reason to nest a module? Also we're already in a `#[cfg(test)]` context here.

florian-bellotti: Yes, it just a way to structure tests by modules. But I can change that

PR #777, opened 2025-08-25
chore: add Clone trait to PoseidonHasher

Yukigaru:
Added `Clone` trait to `PoseidonHasher` in order to use pre-computed hashes in software that uses Starknet.
For example, in my case, I'm hashing an order and it emplies four hash operations (PoseidonHasher::update calls), in which the first two (message_felt and domain hash) are immutable and the same for any order. I'd like to hash those fields once, save the hasher and then clone it to continue with the rest two fields.

I wrote and run a test for that, but haven't included it since it seems obvious.

cargo fmt and clippy: checked.

Thanks!


xJonathanLEI (approved):
LGTM thanks!

PR #781, opened 2025-09-17
chore: update `starknet-types-core` to 0.3.0

glihm:
There is a breaking change between `0.2.0` and `0.2.1` where `ProjectivePoint` `new` method is now returning a `Result`.

In the proposed changes, `new_unchecked` is used since those values aim to be constant and they are known.

A display format has been changed due to some refactoring upstream to follow `std::fmt` spec.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.24.1"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f5fb1d8e4442bd405fdfd1dacb42792696b0cf9cb15882e5d097b742a676d375"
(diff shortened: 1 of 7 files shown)


glihm:
Should we have a nightly version fixed instead of taking the latest one? Or it's recommended to keep this way, and updating dependencies to ensure they build on nightly?

xJonathanLEI:
IMO they should just yank `0.2.0`? Also what we have today doesn't stop people from using `0.2.1` or higher right?

0xLucqs:
you can update to `0.3.0` it's the same as `0.2.1` but with proper semver and `0.2.1` was yanked

snow-avocado:
Highly appreciate if you can update to 0.3.0 starknet-types-core due to the issue noted at https://github.com/starknet-io/types-rs/issues/135. Specifically compilation fails on size-of under rust 1.90.0.

glihm:
bump @xJonathanLEI

glihm:
Will close, not sure this PR is actually relevant based on the discussion in #782.

PR #782, opened 2025-10-08
Bump `types-rs` to 0.2.3

gabrielbosio:


dorimedini-starkware (approved):

xJonathanLEI:
I'm not sure about this. What's the rationale for the bump? Is there a security or other critical issues with `<0.2.3`? As a library crate we should stay as forgiving as possible in terms of crate version ranges.

gabrielbosio:
@xJonathanLEI I understand that it makes sense to keep `types-rs` in 0.2.0 but in this case we had to fix a performance regression related to Pedersen hashing (~120%) in 0.2 which introduced a breaking change. Given the context we had to make an exception regarding semver, that's why these changes were introduced in 0.2.3.

dorimedini-starkware:
@gabrielbosio @xJonathanLEI I am fine with merging this or closing this without merging, as long as we have some 0.2.x version of types-rs we can compile with the 0.13.0 lambdaworks

PR #790, opened 2026-02-04
Fix stark curve modulus

FilipLaurentiu:
According to https://github.com/xJonathanLEI/starknet-rs/blob/a9895554ce67ed3547a262728d18aba68451df7a/starknet-curve/src/curve_params.rs#L5 the Stark Curve order is `0x0800000000000010ffffffffffffffffb781126dcae7b2321e66a241adc64d2f`

`0x0800000000000011000000000000000000000000000000000000000000000001` is the field modulus


FilipLaurentiu:
https://docs.starknet.io/learn/protocol/cryptography#the-stark-curve

